PR #23, opened 2019-10-15
Implement Cascadia Basin seawater.

kkrings:
This is a new seawater implementation for Cascadia Basin, which is located in the Northeast Pacific Ocean at a depth of 2.66 km. Cascadia Basin is a potential site for future neutrino telescopes.

### DIFF
--- a/private/PROPOSAL/medium/Medium.cxx
+++ b/private/PROPOSAL/medium/Medium.cxx
@@ -710,6 +710,33 @@ AntaresWater::AntaresWater(double rho)
 
 }
 
+// chemical composition is normalized with respect to hydrogen: calculate a
+// total atomic weight that yields two hydrogen atoms; use this weight to
+// calculate the numbers of atoms for all other elements
+CascadiaBasinWater::CascadiaBasinWater(double rho)
+    : Medium(
+        "cascadiabasinwater",
+        rho,
+        75.,       // I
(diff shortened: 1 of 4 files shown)


private/PROPOSAL/medium/Medium.cxx
```diff
@@ -710,6 +710,33 @@ AntaresWater::AntaresWater(double rho)
...
+        75.,       // I
+        -3.5017,   // C
+        0.09116,   // a
+        3.4773,    // m
+        0.2400,    // X0
+        2.8004,    // X1
+        0.,        // d0
+        1.0400322, // massDensity
+        {
+            std::shared_ptr<Components::Component>(new Components::Hydrogen(2.0)),
```
MaxNoe: Maybe make_shared would be appropriate here?

https://en.cppreference.com/w/cpp/memory/shared_ptr/make_shared

kkrings: Yes, makes sense. I've to admit that I simply copied the Antares water and then implemented my changes.

kkrings: Can I simply implement the proposed changes, commit to my fork, and the pull request gets automatically updated?

MaxNoe: Yes

PR #28, opened 2019-11-04
Avoid pulling random numbers at scattering when they are not needed.

MaxSac:
Scattering will now return a pair of averaged scattering direction
and the direction of the particle at the end of propagation step.
In "AdvanceParticle" scattering (direction change) and displacement will
from now on consider separately.

The testfile must be modified, since scattering no longer takes over
the displacement of the particle.

### DIFF
--- a/private/PROPOSAL/Sector.cxx
+++ b/private/PROPOSAL/Sector.cxx
@@ -630,9 +630,9 @@ void Sector::AdvanceParticle(double dr, double ei, double ef) {
     Vector3D n_i(particle_.GetDirection());
 
     if( sector_def_.scattering_model != ScatteringFactory::Enum::NoScattering ){
-        std::shared_ptr<std::pair<Vector3D, Vector3D>> directions = scattering_->Scatter(dr, ei, ef);
-        u = directions->first;
-        n_i = directions->second;
+        std::shared_ptr<Directions> directions = scattering_->Scatter(dr, ei, ef);
+        u = directions->u_;
+        n_i = directions->n_i_;
     }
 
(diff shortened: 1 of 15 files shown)


private/PROPOSAL/Sector.cxx
```diff
@@ -625,7 +625,21 @@ void Sector::AdvanceParticle(double dr, double ei, double ef) {
...
-    scattering_->Scatter(dr, ei, ef);
+    std::pair<Vector3D, Vector3D> directions; // u, n_f
+
+    if( sector_def_.scattering_model == ScatteringFactory::Enum::NoScattering ){
+        directions.first = particle_.GetDirection();
+        directions.second = particle_.GetDirection();
+        double rnd1 = RandomGenerator::Get().RandomDouble();
+        double rnd2 = RandomGenerator::Get().RandomDouble();
+        double rnd3 = RandomGenerator::Get().RandomDouble();
+        double rnd4 = RandomGenerator::Get().RandomDouble();
```
MaxSac: Just to reproduce the test. In general these random numbers should be avoided.

private/PROPOSAL/scattering/Scattering.cxx
```diff
@@ -20,12 +22,14 @@ using namespace PROPOSAL;
  ******************************************************************************/
 
 Scattering::Scattering(Particle& particle)
-    : particle_(particle)
+    : particle_(particle),
+      directions_(Vector3D(), Vector3D())
```
sudojan: Why is this a private variable?

PR #29, opened 2019-11-04
WIP: Submodules

MaxNoe:
Using submodules for the dependencies. Just for thinking for now. Might make things more complicated.

Alternatively, one could use the `ExternalDependency` extension.


.gitmodules
```diff
@@ -0,0 +1,9 @@
+[submodule "extern/pybind11"]
+	path = extern/pybind11
+	url = https://github.com/pybind/pybind11
+[submodule "extern/log4cplus"]
+	path = extern/log4cplus
+	url = https://github.com/log4cplus/log4cplus
```
Jean1995: specify version?

MaxNoe: See further down. Submodules point to a specific commit. I used the 2.0.4 tag.

To update use: 
```
$ cd extern/log4cplus
$ git checkout <tag>
$ cd ../..
$ git add log4cplus
$ git commit -m 'Update log3cplus to version <tag>'
```

Issue #26, opened 2019-10-31
Segmentation fault when propagating decay particles


daniel-zid:
I've had a strange error. In my Python code, I create a dictionary containing propagators. Using the propagators in this dictionary, I propagate either taus or antitaus. If one of the decay products is a muon, I pass the properties of this muon to the correspondent propagator and then propagate it to see if it creates particle showers. The weird thing is that this propagation randomly gives segmentation faults.

I can provide you with a simple code that reproduces this error.

Jean1995:
I encountered similar problems when trying to create showers with PROPOSAL using python-bindings.

It is important to know that the return values of a propagator, the list of secondaries, are actually references to objects. These objects are managed by an instance of Output (which is shared by all propagators in your dictionary). Every time a new particle is propagated these objects are deleted so that the new list of secondaries can be saved. I believe that refering to a particle that has already been deleted could be causing the segfaults.

I was able to deal with this problem by directly storing all information from the secondaries that I need in python (before a new particle is propagated and therefore before the information is deleted).

MaxNoe:
Shouldn't this be solved by always using shared pointers? pybind11 advertises that these "just work™"

daniel-zid:
That makes sense. It seems that if I don't propagate while iterating the
array of secondary particles,
everything works. Thank you!

sudojan:
This should now be fixed.

PR #30, opened 2019-11-07
Fix for Issue #26

MaxSac:
Return Secondaries as shared Pointer. Maybe Secondaries should get some utility functions.
Secondaries are pybinded. To get the vector of dynamicData, "secondaries.particles" has to
be called.


MaxSac:
Memory fix cause another unattractive problem. The garbage collector for each `secondaries` is only called when the program is terminated. That makes it slow and consumes a lot of memory.
```
import pyPROPOSAL as pp

prop = pp.Propagator( particle_def=pp.particle.MuMinusDef.get(), config_file="PATH_TO_CONFIG/config.json")
mu = prop.particle
mu_init = pp.particle.Particle(mu)
mu_init.energy = 1e9
mu_backup.direction = pp.Vector3d(0,0,-1)

for i in range(1000):
    mu.inject_state(mu_backup)
    secondaries = prop.propagate(1e6)
    print(secodaries.particles[0].id)
```
@MaxNoe is there a way to avoid the problem, than explicitly calling the garbage can collector?

MaxNoe:
For loops in python don't have scope. Functions do. So if you use a function to propagate a single particle, the secondaries should be cleared up. Or just using `del secondaries`.

You can check memory consumption using psutils;

```
process = psutil.Process(os.getpid())
print(f'Currently using {p.memory_info().rss / 1024**2:.1f} MB RAM')
```

MaxNoe:
One question I have: why is this a shared pointer on a new custom class that basically just wraps a vector? why not `std::vector<shared_ptr<Particle>>` or what ever would be appropriate?

That would have the advantage, that you do not have to reimplement all the methods and protocols of vector. It iteration won't work currently.

MaxNoe:
Sorry, wrong button

public/PROPOSAL/Secondaries.h
```diff
@@ -0,0 +1,56 @@
...
+#pragma once
+
+
+#include "PROPOSAL/particle/Particle.h"
+#include <memory>
+#include <vector>
+
+namespace PROPOSAL {
+
+class Secondaries {
```
MaxNoe: Why this wrapper class? Why not just use the `std::vector<std::shared_ptr<DynamicData>>` directly?

MaxSac:
> One question I have: why is this a shared pointer on a new custom class that basically just wraps a vector? why not `std::vector<shared_ptr<Particle>>` or what ever would be appropriate?
> 
> That would have the advantage, that you do not have to reimplement all the methods and protocols of vector. It iteration won't work currently.

There should be some querys on the wrapper class which are not yet implemented.
The pointer itself is needless and i will delete it in next commit.

public/PROPOSAL/particle/ParticleDef.h
```diff
@@ -332,29 +332,29 @@ PROPOSAL_MAKE_HASHABLE(PROPOSAL::ParticleDef, t.mass, t.lifetime, t.charge)
 namespace PROPOSAL {
 static std::map<const int, const ParticleDef&> Id_Particle_Map
```
MaxNoe: wouln't `const ParticleType` as type for the map be better than all those casts to int?

PR #34, opened 2020-02-17
Use smart pointer in Propagator and Sector

MaxSac:
feature: Use smart pointer to reducing memory leaks and avoid explicit clone operations.
fix: misspelling ParseParameterConifg


private/PROPOSAL/Propagator.cxx
```diff
@@ -801,13 +799,27 @@ Geometry* Propagator::ParseGeometryConifg(const std::string& json_object_str)
...
+        log_debug("The 'hierarchy' option is not set. Use default (0)");
+    }
+
     // --------------------------------------------------------------------- //
     // Check type of geometry and set members
     // --------------------------------------------------------------------- //
-
-    if (PROPOSAL::Sphere* sphere = dynamic_cast<PROPOSAL::Sphere*>(geometry)) {
+    
+    if (geometry->GetName() == "Sphere") {
```
MaxNoe: This looks like it should be a switch and name an enum?

private/PROPOSAL/Sector.cxx
```diff
@@ -45,7 +45,7 @@ Sector::Definition::Definition()
     , location(Sector::ParticleLocation::InsideDetector)
     , utility_def()
     , cut_settings()
-    , medium_(new Ice())
+    , medium_(std::make_shared<const Medium>(Ice()))
```
MaxNoe: This is stack allocating a Ice and passing this to the copy constructor of `Medium`.

This should be `std::make_shared<const Ice>()`

private/PROPOSAL/Propagator.cxx
MaxSac: Thank you, I was not aware of the implicit conversion in smart pointers.

private/PROPOSAL/Sector.cxx
MaxSac: Thank you, I was not aware of the implicit conversion in smart pointers.

private/PROPOSAL/Propagator.cxx
MaxSac: I think the problem is solved by commit b98a122 .?

MaxSac:
> Does this just silently do nothing if config is not what was expected?

Condition checks are in the constructor.

MaxSac:
> Does this just silently do nothing if config is not what was expected?

Fixed in 06e239f .

private/PROPOSAL/geometry/Geometry.cxx
```diff
@@ -73,6 +73,7 @@ Geometry::Geometry(const nlohmann::json& config)
 
     if(not config.contains("origin"))
         throw std::invalid_argument("No geometry originfound.");
+    position = Vector3D(config.at("origin");
```
MaxNoe: missing parenthese??

MaxSac: and underscore too :sweat:

private/PROPOSAL/Sector.cxx
```diff
@@ -62,11 +64,66 @@ Sector::Definition::Definition(const Definition& def)
...
+            EnergyCutSettings cut_settings(config.at(cut.first));
+            location = cut.second;
+            break;
+        }
+    }
+
+    std::string scattering_model_str = config.value("scattering", "highlandintegral");
+    scattering_model = ScatteringFactory::Get().GetEnumFromString(scattering_model_str);
+
+    if(config.contains("shape")) {
```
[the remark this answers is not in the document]
MaxSac: Done in 3767699

private/PROPOSAL/Sector.cxx
```diff
@@ -62,11 +64,66 @@ Sector::Definition::Definition(const Definition& def)
...
+        std::string shape = config["geometry"]["shape"];
+        if (shape == "sphere") {
+            geometry_ = std::make_shared<const Sphere>(config["geometry"]);
+        } else if (shape == "box") {
+            geometry_ = std::make_shared<const Box>(config["geometry"]);
+        } else if (shape == "cylinder") {
+            geometry_ = std::make_shared<const Cylinder>(config["geometry"]);
+        } else {
+            throw std::invalid_argument("You need to specify a detector for each sector");
```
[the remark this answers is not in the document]
MaxSac: Done in 3767699

private/PROPOSAL/Sector.cxx
```diff
@@ -62,11 +64,66 @@ Sector::Definition::Definition(const Definition& def)
...
+        }
+    }
+
+    std::string medium_name = "water";
+    double density_correction = 1.0;
+    if(config.contains("medium"))
+    {
+        config.at("medium").get_to(medium_name);
```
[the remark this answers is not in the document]
MaxSac: solved in fa6dc8c

private/PROPOSAL/Sector.cxx
```diff
@@ -62,11 +64,71 @@ Sector::Definition::Definition(const Definition& def)
...
+        std::make_pair("cuts_infront", Sector::ParticleLocation::InfrontDetector),
+        std::make_pair("cuts_inside",Sector::ParticleLocation::InsideDetector),
+        std::make_pair("cuts_behind",Sector::ParticleLocation::BehindDetector),
+    };
+
+    for (const auto& cut : cuts) {
+        if(config.contains(cut.first)) {
+            EnergyCutSettings cut_settings(config.at(cut.first));
+            location = cut.second;
+            break;
```
MaxNoe: why break here?  and why is nothing done with `cut_settings`. this goes out of scope after the if.

MaxSac: Energy Cuts has three possible keywords. The keyword itself contains information about the position. 
Construction of `cut_settings` is specified in 3082146. It's a private class variable which break underscores variable definition convention.

PR #62, opened 2020-02-26
Fix sectors unittest bug because newton raphson returns zero if perci…

MaxSac:
…sion is not reached.

Now it warns and give best guess back.


private/PROPOSAL/propagation_utility/PropagationUtilityInterpolant.cxx
```diff
@@ -221,12 +221,8 @@ double UtilityInterpolantDisplacement::GetUpperLimit(double ei, double rnd) {
     df = [&](double ef) { return interpolant_diff_->Interpolate(ef); };
 
     int MaxSteps = 200;
-    try {
-        return std::max(NewtonRaphson(f, df, 0, ei, ei, MaxSteps,
-                    1.e-6), utility_.GetParticleDef().mass);
-    } catch (MathException& e) {
-        return utility_.GetParticleDef().mass;
```
Jean1995: This case should still be catched! If we want to solve this displacement integral for distances where to solution can be found, not catching the exception will cause an error (and therefore, the PropagationTest do fail)

PR #66, opened 2020-03-02
add binary operator to vector for less redundand code

MaxSac:


private/PROPOSAL/math/Vector3D.cxx
```diff
@@ -106,6 +106,11 @@ void Vector3D::swap(Vector3D& vector_3d)
     // swap(cylindric_height_,  vector_3d.cylindric_height_);
 }
 
+Vector3D Vector3D::operator+=(const Vector3D& rhs)
```
MaxNoe: This does not mutate `this`, does it? Not what I would expect from `+=`.

Herb Sutte makes the strong point, that `operator+` should always be implemented via `operator+=`, like this:

```
T& T::operator+=( const T& other ) {
    //...
    return *this;
}

T operator+( T a, const T& b ) {
    a += b;
    return a;
}
```

https://herbsutter.com/2013/05/20/gotw-4-class-mechanics/
This `operator+=` does not even mutate `*this`, right?

public/PROPOSAL/math/Vector3D.hpp
```diff
@@ -0,0 +1,216 @@
...
+using std::sin;
+
+namespace PROPOSAL {
+template <typename T> class Vector3D {
+protected:
+    std::array<T, 3> values;
+
+public:
+    Vector3D(array<T, 3> val)
+        : values(val){};
```
MaxNoe: std::move val here

public/PROPOSAL/math/Vector3D.hpp
```diff
@@ -0,0 +1,216 @@
...
+
+    T magnitude() const override;
+    Cartesian3D<T> ToCartesian() const;
+};
+} // namespace PROPOSAL
+
+namespace PROPOSAL {
+template <typename T> bool Vector3D<T>::operator==(const Vector3D&& rhs) const
+{
+    for (auto i : indices(values)) {
```
MaxNoe: const auto&

public/PROPOSAL/math/Vector3D.hpp
```diff
@@ -0,0 +1,216 @@
...
+{
+    for (size_t i = 0; i < 3; ++i)
+        l_vec[i] -= r_vec[i];
+    return l_vec;
+}
+
+template <typename T>
+Cartesian3D<T> Cartesian3D<T>::operator+=(const Cartesian3D<T>& rhs)
+{
+    return *this + rhs;
```
MaxNoe: same error as last time! 

operator+= should mutate in place and return a reference. This creates unecessary copies.

PR #67, opened 2020-03-02
Fix some compiler warnings about not defined here

MaxNoe:


public/PROPOSAL/math/Function.h
```diff
@@ -44,7 +44,7 @@ class Polynom {
     Polynom(const Polynom&);
 
     ~Polynom() {};
-    // Polynom& operator=(const Polynom& poly);
+    Polynom& operator=(const Polynom& poly) = default;
```
Jean1995: https://stackoverflow.com/questions/51863588/warning-definition-of-implicit-copy-constructor-is-deprecated

Good explanation

Issue #63, opened 2020-02-26
Use cmakes builtin mechanisms for build types


MaxNoe:
cmake has a builtin mechanism for building types and managing cpp flags for them.

We should use that instead of the current `option DEBUG`

MaxNoe:
done in #68

Issue #11, opened 2018-09-20
Require a more modern cmake version?


MaxNoe:
I think the CMake files could get a lot more readable and a lot shorter using a modern cmake version, at least something > 3, see e.g. this post:

https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1

MaxNoe:
Done in #68

PR #68, opened 2020-03-03
Refactor CMake, fixes #64, #63, #11

### LINKED ISSUE #64: Use cmake properties instead of setting cxx flags
It's considered bad style to explicitly set compiler arguments such as `--std=c++11` instead of using cmake options such as:

```
target_compile_features(PROPOSAL PUBLIC cxx_std_11)
set_target_properties(PROPOSAL PROPERTIES CXX_EXTENSIONS OFF)
```

we should switch to this.

MaxNoe:


sudojan:
which minimal cmake version is really required?
3.10 as in `CMakeLists.txt`
or 3.8 as in `cmake/Standalone_CMakeLists.cmake`

MaxNoe:
Seems to be 3.8, worked fine and lower is not possible as this version introduced support for the c++ standards https://cmake.org/cmake/help/v3.17/release/3.8.html

MaxNoe:
I set it consistently now.

MaxNoe:
And I also download cmake 3.8.2 on travis and use that so that we introduce changes that require a newer cmake version, the tests will break.

PR #74, opened 2020-03-06
remove clang warnings in medium

sudojan:
with the latest PR the compile log is full of the clang warning
`warning: 'PROPOSAL::Salt::create' hides overloaded virtual function [-Woverloaded-virtual]`
for every Medium
used the solution suggested here
https://stackoverflow.com/questions/18515183

### DIFF
--- a/tests/Propagation_TEST.cxx
+++ b/tests/Propagation_TEST.cxx
@@ -12,7 +12,7 @@ TEST(Comparison, Comparison_equal)
 {
     Sector::Definition sector_def;
     sector_def.location = Sector::ParticleLocation::InsideDetector;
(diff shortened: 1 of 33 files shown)


tests/Propagation_TEST.cxx
```diff
@@ -12,7 +12,7 @@ TEST(Comparison, Comparison_equal)
 {
     Sector::Definition sector_def;
     sector_def.location = Sector::ParticleLocation::InsideDetector;
-    sector_def.SetMedium(std::make_shared<Medium>(Water()));
+    auto medium = std::make_shared<const Water>();
```
sudojan: shouldn't this line be
`sector_def.SetMedium(std::make_shared<const Water>());`

PR #76, opened 2020-03-12
Example scripts

sudojan:
Adapt the python scripts to build the TestFiles and the example scripts to the new version.


private/PROPOSAL/Propagator.cxx
```diff
@@ -151,85 +152,117 @@ Propagator::Propagator(
...
+            sec_def_global->location = Sector::ParticleLocation::InsideDetector;
+        }
+
+        std::string scattering_model_str
+            = json_global.value("scattering", "highlandintegral");
+        sec_def_global->scattering_model
+            = ScatteringFactory::Get().GetEnumFromString(scattering_model_str);
+        sec_def_global->do_stochastic_loss_weighting
+            = json_global.value("stochastic_loss_weighting", false);
+        sec_def_global->stochastic_loss_weighting
```
sudojan: asking two times for "stochastic_loss_weighting" seems to be a copy-paste-mistake

private/PROPOSAL/Propagator.cxx
```diff
@@ -151,85 +152,117 @@ Propagator::Propagator(
...
         uniform = json_global.value("uniform", uniform_);
 
-        sec_def_global.reset(new Sector::Definition(json_global));
+
+        if(json_global.contains("cuts")){
+            sec_def_global->cut_settings = EnergyCutSettings(json_global.at("cuts"));
+            sec_def_global->location = Sector::ParticleLocation::InsideDetector;
+        }
+
+        std::string scattering_model_str
```
sudojan: There should be one place for the default values of the SectorDef

private/PROPOSAL/Propagator.cxx
MaxSac: Done in e09cb5f

private/Interface/python/particle.cxx
```diff
@@ -269,7 +269,16 @@ void init_particle(py::module& m) {
         .def_property_readonly("parent_particle_energy", &Secondaries::GetParentParticleEnergy)
         .def_property_readonly("energy", &Secondaries::GetEnergy)
         .def_property_readonly("time", &Secondaries::GetTime)
-        .def_property_readonly("propagated_distance", &Secondaries::GetPropagatedDistance);
+        .def_property_readonly("propagated_distance", &Secondaries::GetPropagatedDistance)
+        .def_property("entry_point",
+                    &Secondaries::GetEntryPoint,
+                    &Secondaries::SetEntryPoint)
```
MaxSac: Why should there be an interface to set entry and exit point by user?

resources/config.json
```diff
@@ -26,8 +26,15 @@
 		"photo_multiplier" : 1,
 		"ioniz_multiplier" : 1,
 		"epair_multiplier" : 1,
+        "ioniz": "IonizBetheBlochRossi",
+        "epair": "EpairKelnerKokoulinPetrukhin",
 		"brems" : "BremsKelnerKokoulinPetrukhin",
 		"photo" : "PhotoAbramowiczLevinLevyMaor97",
+        "annihilation": "none",
```
MaxSac: None is the default value. Why set it explicit?

resources/examples/standalone/plot_energy_probability.py
```diff
@@ -7,7 +7,7 @@
 from matplotlib.colors import LogNorm
 
 import numpy as np
-
+from tqdm import tqdm
```
MaxSac: Does proposal have a dependency on `tqdm`?

resources/examples/standalone/plot_entry_exit_closest_approach.py
```diff
@@ -81,10 +81,11 @@ def main():
     ]
 
     for jdx in range(len(start_energies)):
```
MaxSac: Nested loop is more elegant, but not important here.

resources/examples/standalone/plot_range.py
```diff
@@ -1,5 +1,6 @@
 import pyPROPOSAL as pp
 import matplotlib.pyplot as plt
+from tqdm import tqdm
```
MaxSac: the same as above

resources/examples/standalone/plot_energy_probability.py
MaxNoe: I would not say that examples contribute to "dependencies", PROPOSAL does not depend on matplotlib or numpy either.

I would also remove these "try except ImportError" lines, `ImportError no module named matplotlib` will be pretty clear to every python user, and not really worse that `ImportError you need to install matplotlib`

private/Interface/python/particle.cxx
sudojan: Thats correct; just took it from the c lib.
This can both be deleted.

resources/config.json
sudojan: To show, what possibilities the user has.

PR #80, opened 2020-03-15
Add setup, fixes #42

### LINKED ISSUE #42: Easier PROPOSAL installation
Installing PROPOSAL, e.g. via pip install or an AUR package for arch users would be favorable.

MaxNoe:
Requires #79 review and discuss that first.

Actually this is the main reason for #79.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,146 @@ cmake_uninstall.cmake
 *.pdf
 *.swp
 tests/TestFiles*.tar.gz
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+pip-wheel-metadata/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -12,3 +12,146 @@ cmake_uninstall.cmake
...
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
```
sudojan: build/ already exist in `.gitignore`

setup.py
```diff
@@ -0,0 +1,96 @@
...
+    raise OSError('You need cmake >= 3.8')
+
+
+class CMakeExtension(Extension):
+    def __init__(self, name, source_dir=None, target=None, **kwargs):
+        if source_dir is None:
+            self.source_dir = SETUP_DIR
+        else:
+            self.source_dir = os.path.join(SETUP_DIR, source_dir)
+        self.target = None
```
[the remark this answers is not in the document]
MaxNoe: ah right, sorry.

We don't use it here anyways.

.gitignore
MaxNoe: I just append the standard python gitignore

sudojan:
it's working with `python setup.py install`
unfortunately, this is not the ideal solution, because uninstalling is manually and not as easy as with pip
When installing with `pip install .` it builds pyPROPOSAL successfully.
But during `import pyPROPOSAL` the following Error occurs:
`ImportError: dlopen(…/pyPROPOSAL.cpython-37m-darwin.so, 2): Library not loaded: @rpath/libPROPOSAL.dylib`
`Referenced from: …darwin.so`
`Reason: image not found`

PR #81, opened 2020-03-16
Fix interpolation pointers

Jean1995:
The Interpolation issue caused by undefined unique_ptr objects has been solved. Now, the InitializeInterpolation function returns a unique_ptr instead of accepting and resetting one (this led to an undefined behaviour)

### DIFF
--- a/private/test/example.cxx
+++ b/private/test/example.cxx
@@ -1,6 +1,5 @@
 #include "PROPOSAL/PROPOSAL.h"
 #include <vector>
-#include <iostream>
 
 using namespace PROPOSAL;
 
--- a/private/test/example.cxx
+++ b/private/test/example.cxx
@@ -1,80 +1,26 @@
-
-#include <fstream>
-#include <iostream>
-
 #include "PROPOSAL/PROPOSAL.h"
+#include <vector>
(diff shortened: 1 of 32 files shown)


private/test/example.cxx
```diff
@@ -1,80 +1,26 @@
-
-#include <fstream>
-#include <iostream>
-
 #include "PROPOSAL/PROPOSAL.h"
+#include <vector>
+#include <iostream>
```
MaxSac: artefact

public/PROPOSAL/methods.h
```diff
@@ -159,7 +159,7 @@ bool FileExist(const std::string path);
 // ----------------------------------------------------------------------------
 std::string Centered(int width, const std::string& str, char fill = '=');
 
-typedef std::vector<std::pair<InterpolantBuilder*, std::unique_ptr<Interpolant>> > InterpolantBuilderContainer;
+typedef std::vector<InterpolantBuilder*> InterpolantBuilderContainer;
```
MaxSac: `using InterpolantBuilderContainer = std::vector<InterpolantBuilder*>`

PR #83, opened 2020-03-16
Export proposal

MaxNoe:
This exports `lib/cmake/PROPOSALConfig.cmake` and `lib/cmake/PROPOSALConfigVersion.cmake` so that other cmake projects can do `find_package(PROPOSAL)`.

See https://youtu.be/bsXLMQ6WgIk?t=2270
and https://cliutils.gitlab.io/modern-cmake/chapters/install/installing.html


cmake/Standalone_CMakeLists.cmake
```diff
@@ -20,8 +20,9 @@ SET(CMAKE_SKIP_BUILD_RPATH  FALSE)
 
 # when building, don't use the install RPATH already
 # (but later on when installing)
-SET(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
-SET(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
```
sudojan: isn't using `CMAKE_INSTALL_LIBDIR` more elegant

PR #84, opened 2020-03-17
Test python

MaxNoe:
First step to test the python bindings, adds a matrix element with language python 3.6 that does the pip install.

I also found this python script that looked like a test and converted it so pytest can actually run it. It passes... @sudojan is that a meaningful test? I mean it at leasts propagates some particles so it at leasts does a api regression test...


sudojan:
It's more a test if the continuous losses are calculated or retrieved correctly.
But yeah, in principle for such a test needed here, one has to propagate just a couple of muons and make sure the output consists of the interaction points.
So maybe the checks at the end are not really necessary for a quick test, but also not time-consuming.

Maybe one should more make it clear that the path to the interpolation tables should exist, otherwise the tables are stored in memory.

MaxNoe:
Ok,  but then let's merge this and design some meaningful python tests in another PR

MaxNoe:
with this, the actual next step could be to deploy to pypi.

sudojan:
for the python build these nearly 2000 lines of output arise, telling e.g.

Skipping link https://files.pythonhosted.org/packages/38/5c/854a6aa92f497b1a35df560500bdf25a169e79055945f15146c7a8c25ebc/cmake-0.7.0-cp35-cp35m-win_amd64.whl#sha256=b2756c2c37aeb03921e0eea85abfafa780aac1452e5d2ab9303a32dc4b82fa75 (from https://pypi.org/simple/cmake/); it is not compatible with this Python

MaxNoe:
This comes from the `-v` option to pip, which I used to show the build output of pip install.
I'll remove it.

PR #87, opened 2020-03-18
Initialize log4cplus correctly and allow setting config via env, fixes #86

### LINKED ISSUE #86: Installed python package does not find log4cplus config
```
log4cplus:ERROR could not open file /tmp/pip-req-build-79oqjkru/resources/log4cplus.conf
log4cplus:ERROR No appenders could be found for logger (PROPOSAL).
log4cplus:ERROR Please initialize the log4cplus system properly.
```

The path is the temporary build path.

Any ideas how to adapt this, so it can find a path at installation?

MaxNoe:
* Properly use log4cplus::Initializer to setup log4cplus
* Load a basic config
* Add the option to provide the log config for PROPOSAL via an env variable `PROPOSAL_LOG_CONFIG`
* remove build time path for log config


sudojan:
The `resources/log4cplus.conf` seems to be obsolete now.
Or should it stay as an example?

public/PROPOSAL/Logging.h
```diff
@@ -51,15 +53,28 @@ class Logging
...
-        log4cplus::PropertyConfigurator::doConfigure(LOG4CPLUS_TEXT(LOG4CPLUS_CONFIG));
+        log4cplus::BasicConfigurator basic_config;
+        basic_config.configure();
+
         logger = log4cplus::Logger::getInstance(LOG4CPLUS_TEXT("PROPOSAL"));
+        logger.setLogLevel(log4cplus::INFO_LOG_LEVEL);
+
+        // support setting the log4cplus config via an environment variable
+        const char* config = std::getenv("PROPOSAL_LOG_CONFIG");
+        if (config != NULL) {
```
MaxSac: maybe `if (config) { }` ?

MaxSac: or `if (const char* config = std::getenv("PROPOSAL_LOG_CONFIG"))` ?

sudojan:
> The `resources/log4cplus.conf` seems to be obsolete now.
> Or should it stay as an example?

better leave it there if people wants to set the `PROPOSAL_LOG_CONFIG`

sudojan:
or rename it to `log4cplus_example.conf`

public/PROPOSAL/Logging.h
MaxNoe: I don't really like PEP 572 :rofl:

MaxNoe:
It doesn't really serve as a meaningful example for PROPOSAL, as PROPOSAL only uses a single logger.
Usually you use loggers per module or even function.

I'll remove it

PR #93, opened 2020-03-21
Fix pypi deploy

MaxNoe:
Pypi deploy failed because travis was in the wrong directory after building the tar ball for the github deploy.


MaxNoe:
The full links there so the people can click on them in the pypi project description:

https://pypi.org/project/proposal

PR #94, opened 2020-03-21
fix position-direction-mix-up in python interface

sudojan:


MaxNoe:
Let's directly bump the version here and make the new release then?

PR #95, opened 2020-03-21
prepare for release 6.1.1

sudojan:


MaxNoe:
The cmake example in the readme should also use find_package now. Or better remove the duplication and link the install page

PR #103, opened 2020-06-23
Add missing header includes

MaxNoe:
Missing headers, needed to make it compile with gcc 10.


8me:
What is the current status of this PR, @sudojan?

Issue #104, opened 2020-07-22
Secondaries Terminology Clarifications


wjwoodley:
What exactly is meant by the "secondaries"? What is `sec.energy`, and why does it start nearly at the primary particle energy and then continuously decrease?

It seems most of the information in `sec` is a copy of the information in `sec.particles` (for example, `sec.energy == [i.energy for i in sec.particles]` returns `True`). Is there information about the propagating particle stored anywhere?

And what is meant by parent particle? Is there a relationship between `sec.energy` and `sec.parent_particle_energy` (for example, `sec.energy[i] == sec.parent_particle_energy[i + 1]` returns `False`)?

sudojan:
Hi,

the output is indeed unintuitive and has changed from version 5 to 6 as first restructuring arrangements were included in this release. (The rest of the restructuring is in the current branch `restructure_parametrization`)
The energy of the Secondary is the energy of the muon after a propagation step. This unintuitive Output is due to the restructuring of PROPOSAL for CORSIKA, where it serves as the propagator for the em interactions. Therefore the output has changed to a Track with the muon properties creating secondaries.
This unintuitive behavior will be corrected with the next big release when the `restructure_parametrization` branch is merged. Then one can decide to get either the energy losses or the muon track.

`sec.energy` is a shortcut for getting the energies at the interaction points (in C++ these are the `Query` methods). `sec.parent_particle_energy` is the shortcut for the particle energy before the interaction. The reason why `sec.energy[i]` doesn't equal `sec.parent_particle_energy[i+1]` are the continuous losses.

sudojan:
This is improved in MR #137

PR #114, opened 2020-10-25
WIP: Restructuring of Secondaries/Track, Losses and DynamicData

Jean1995:
This merge request for the development branch is intended to improve the return type of the Propagator class, i.e. a Secondaries/Track object, as well as making the objects describing Particles and losses in PROPOSAL more intuitive (see also #104)

### Losses/Particles

At the moment, it is not entirely clear which objects describe energy losses and particle states in PROPOSAL. Therefore, I would like to introduce three different objects:

**1. DynamicData**
_DynamicData_ objects should only be used to described particle states. Examples are the initial state of the Propagator, secondary particles or elements of the Track. DynamicData objects are defined by their ParticleType (PDG code), position, direction, energy, time and propagated distance. I removed most of the Getter/Setter methods to simplify the class.

**2. StochasticLoss**
_Stochastic loss_ objects are used to describe particles losses. This is the object of interest for many users, for example the IceCube experiment. StochasticLoses are described by their (loss)energy, InteractionType, position, direction, time (when the StochasticLoss happend), propagated distance (when the StochasticLoss happend) and parent_particle_energy (initial energy of the particle causing the StochasticLoss).

**3. ContinuousLoss**
As an special loss type, _ContinuousLoss_ objects are introduced. Unlike stochastic losses, continuous losses are not defined by a single energy/position/time and should therefore be treated differently than the StochasticLoss objects. ContinuousLoss objects are destribed by a pair of energies (energies at the beginning and end of the stochastic loss), positions (at the beginning and end of the stochastic loss), times (at the beginning and end of the stochastic loss) and direction(s).

### Track/Secondaries
The output of the Propagator is a class called _Secondaries_ (although I would prefer renaming it, for example to _Track_).
Here, the information of the Propagation process is stored in a vector of _DynamicData_ objects, describing consecutive states of the propagated particle (therefore Track).

From this vector, the class provides the functionality to convert the _DynamicData_ objects into _StochasticLoss_ or _ContinuousLoss_ objects. This requires additional information about the last interaction type of the _DynamicData_ objects. These information are currently stored in the _type_ variable of the _DynamicData_ object, although I believe that this variable should be reserved for the PDG code of the particle object.

Furthermore, the Secondaries/Track class provides additional functionalities, including:
- Calculate the state of a particle when entering/exiting an arbitrary geometry _GetEntryPoint_ and _GetExitPoint_
- Calculate the state of a particle for the closest approach to an arbitrary geometry _GetClosesApproachPoint_
- Queries, filtering requested information from the track (e.g. only Track information inside a certain geometry or only information on specific variables)

### SecondariesCalculator

The SecondariesCalculator provides the functionality to convert energy losses into particle states. An electron-positron pair production for example can be described by a _StochasticLoss_ object. The corresponding SecondariesCalculator turns this _StochasticLoss_ objects into three _DynamicData_ objects: The incident lepton, an electron and and positron.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,6 +156,3 @@ install(
     FILES "${CMAKE_CURRENT_BINARY_DIR}/PROPOSALConfigVersion.cmake"
     DESTINATION lib/cmake/PROPOSAL
 )
-
-add_executable(example example.cxx)
-target_link_libraries(example PRIVATE PROPOSAL)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,3 +156,6 @@ install(
     FILES "${CMAKE_CURRENT_BINARY_DIR}/PROPOSALConfigVersion.cmake"
     DESTINATION lib/cmake/PROPOSAL
 )
+
+add_executable(example example.cxx)
+target_link_libraries(example PRIVATE PROPOSAL)
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ using namespace PROPOSAL;
 int main(){
     ParticleDef mu_def = MuMinusDef::Get();
     Propagator prop(mu_def, "resources/configuration/config.json");
-    DynamicData mu(mu_def.particle_type);
+    ParticleState mu(mu_def.particle_type);
 
     mu.SetEnergy(9e6);
     mu.SetPosition(Vector3D(0, 0, 0))
@@ -216,7 +216,7 @@ prop = pp.Propagator(
 	  config_file="path/to/config.json"
 )
 
-mu = pp.particle.DynamicData(mu_def.particle_type)
+mu = pp.particle.ParticleState(mu_def.particle_type)
 
 mu.energy = 9e6
 mu.direction = pp.Vector3D(0, 0, -1)
--- a/examples/Propagator.ipynb
+++ b/examples/Propagator.ipynb
@@ -53,7 +53,7 @@
     "\n",
     "prop = pp.Propagator(args[\"particle_def\"], [(detector, utility, density_distr)])\n",
     "\n",
-    "initial_state = pp.particle.DynamicData(args[\"particle_def\"].particle_type)\n",
+    "initial_state = pp.particle.ParticleState(args[\"particle_def\"].particle_type)\n",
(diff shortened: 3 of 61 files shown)


maxnoe:
Let me just comment that the name `DynamicData` is very non-descriptive. So if you are now making breaking changes, maybe better come up with something more descriptive?

Also, `ContinuousLoss` should probably be plural, right?

Jean1995:
> Let me just comment that the name `DynamicData` is very non-descriptive. So if you are now making breaking changes, maybe better come up with something more descriptive?

I completely agree, `DynamicData` is still the old name for the object where particle states and losses had no differentiation. I do like the term `ParticleState`, since we already have the term `ParticleDef` for all the "non-dynamic" atributes of particles (mass, lifetime, etc.).



> Also, `ContinuousLoss` should probably be plural, right?

One `ContinousLoss` object describes the continuous loss(es) between two positions. I am not entirely sure whether this should seen as "one accumulated continuous loss" (singular) or "the accumulation of continuous energy losses" (plural). 


> Why does a loss have a direction? And what about daughter particles of a `StochasticLoss`? (newly created gamma for bremsstrahlung, e⁺ / e⁻ pair for pair production, etc.)?

> Aha! Wouldn't it be better if the `StochasticLoss` already had this particle states, maybe as a `std::vector<std::shared_ptr<Particle>>` ?

Energy losses are highly boosted into one direction, therefore the direction is important for an energy loss (although I agree that this is a tad unintuitive).

The reason why an energy loss is not directly linked to its daughter particles lies in both the way PROPOSAL works at the moment and the different applications:
During the Propagation loop, PROPOSAL only calculates energy losses (e.g.: "My initial particles looses 5% of its energy at this point due to pairproduction). This information is enough for users such as IceCube.
For CORSIKA however, this information is not enough - They are interested in the daughter particles that will end up on the Stack (e.g. "My initial particles creates an electron and a positron with these energies"). 

Therefore, we moved certain calculations (for example the calculation of the asymmetry parameter in pairproduction) to the SecondaryCalculator. These calculations can be done, if necessary, after the propagation process. This will finally convert StochasticLosses to `std::vector<std::shared_ptr<Particle>>`.

Issue #119, opened 2021-01-11
direction value outside the allowed range


pkalaczynski:
There seems to be a problem with how the directional cosines are evaluated. I have a muon track, which before the propagation step in the simulation chain has the directional cosines:
```
dir.x            dir.y            dir.z
-0.200238   -0.046389   -0.978648
```
 but after propagating with PROPOSAL it ends up to be:
```
dir.x            dir.y            dir.z
0.0086624  -0.004045   -1.000020
```

which of course does not make sense, since cos should not give anything smaller than -1.0 ...

I suppose it could be just some precision issue, but I'm not so familiar with the source code, so it's hard for me to judge where could things go wrong.

Jean1995:
Since the direction should be a unit vector, it is indeed wrong that the z component is smaller than -1.0...

Could you provide the energies of the propagation step, i.e. the energies before and after the propagation step when the direction became invalid? Can you also provide the Scattering Model, the energy cut settings and the medium you used?

I have an idea where the origin of this error might be, but I would like to be able to reproduce your error to see if I am right.

pkalaczynski:
PROPOSAL version: 6.1.4

E before propagation: 668.503 GeV
E after propagation:    0.7800031 GeV

Scattering model: HighlandIntegral
Energy cuts: default value used

Medium:
Sea water:
Sea rock:

water composition:
```
  shared_ptr<const PROPOSAL::Medium> med_sw = make_shared<const PROPOSAL::Medium>("gSeaWater",
           1,
           75.0,    // I
           -3.5017, // C
           0.09116, // a
           3.4773,  // m
           0.2400,  // X0
           2.8004,  // X1
           0,       // d0
          RhoSW, // massDensity
           ComponentSW);

RhoSW = 1.03975

ComponentSR.push_back(PROPOSAL::Components::Component(AtomName[iComp].c_str(),Z[iComp],AtomMass[iComp],AtomInMolecule));
```

with the following values

```
H 1 1.00783 2.00048
C 6 12 4.34779e-05
O 8 15.9949 1
Na 11 22.9898 0.00875349
Mg 12 23.985 0.00100372
S 16 31.9721 0.00053035
Cl 17 34.9689 0.0103374
K 19 38.9637 0.00019129
Ca 20 39.9626 0.000186508
Br 35 79.9185 1.56214e-05
```

pkalaczynski:
[...]

PR #120, opened 2021-01-13
Add call to CalculateSphericalCoordinates() at the beginning of Scattering:Scatter

Jean1995:
Add call to CalculateSphericalCoordinates() at the beginning of Scattering:Scatter to ensure that return values of GetTheta() and GetPhi() are well-defined if spherical components have not been calculated yet.

Could be the solution for issue #119

There will be a better, conceptional fix for this issue in version 7.X

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.8)
 
-project(PROPOSAL VERSION 6.1.4 LANGUAGES CXX)
+project(PROPOSAL VERSION 6.1.5 LANGUAGES CXX)
 
 
 IF(I3_PROJECTS)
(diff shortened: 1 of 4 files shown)


Jean1995:
Since UnitTest_Propagation did not correctly initialize the initial direction of the particle, this PR caused the UnitTests to fail. Therefore, I had to rebuild the corresponding TestFile.

As a cross check, I successfully verified that this PR only influenced the particle positions and directions inside the UnitTests (and not the energies or interaction types, since the medium is homogeneous).

Furthermore, I compared the final particle positions from the old and new TestFiles:

image

Apparently, there is no systematic difference (as expected) comparing the old and new TestFiles (I increased the statistics by a factor 10 to create this plot).

Therefore, I believe it was justified to rebuild Propagation_TestFile 😄

sudojan:
This might also be a problem for other parts, where the direction changes e.g. when just calling the decay.

Jean1995:
> This might also be a problem for other parts, where the direction changes e.g. when just calling the decay.

Apparently, Scattering is the only class that uses the Spherical Coordinates of a Vector3D. Decay for example only uses the Cartesian Coordinates.

sudojan (approved):

PR #121, opened 2021-01-14
add deflection parameterizations and sample from exponential

pascalgutjahr:
add parameterization for bremsstrahlung and nuclear interaction using Geant4 manual and pair production of van Ginneken, 1986 (ENERGY LOSS AND ANGULAR CHARACTERISTICSOF HIGH ENERGY ELECTROMAGNETIC PROCESSES)

also added a sampling method from exponential


sudojan:
The Naming could be more descriptive e.g. NaiveBremsstrahlung is not helpful.
Usually in PROPOSAL the name of a cross section Parametrization quote the author(s) of these calculations.
In the case of NaiveBremsstrahlung, you refer to Geant4 and (afaik, however, they don't cite in their manual) they use the calculations of Tsai Rev.Mod.Phys. 46 (1974) 815.

asandrock:
Actually, the distribution f(r) ~ r/(1 + r^2)^2 [Geant4 Physics Reference Manual, somewhere after equation (4.208)] is an approximation for small r to the distribution in [Tsai, equation (3.80)], so they are not equivalent.

src/PROPOSAL/scattering/stochastic_deflection/nuclearInteraction/NaivNuclearInteraction.cxx
```diff
@@ -0,0 +1,33 @@
...
+#include "math.h"
+
+using namespace PROPOSAL;
+
+std::array<double, 2>
+stochastic_deflection::NaivNuclearInteraction::CalculateStochasticDeflection(
+        double e_i, double e_f, std::vector<double> const& rnd) const
+{
+    // ------- NuclInt 
+    auto proton_mass = 938.272; // Proton mass in MeV
```
Jean1995, line 12: For all physical constants, we should use the variables from `PROPOSAL/Constants.h` (in this case, MP for the proton mass) to be consistent throughout the whole project.

sudojan:
> Actually, the distribution f(r) ~ r/(1 + r^2)^2 [Geant4 Physics Reference Manual, somewhere after equation (4.208)] is an approximation for small r to the distribution in [Tsai, equation (3.80)], so they are not equivalent.

then I would still write TsaiApproximation or something like this

src/PROPOSAL/scattering/stochastic_deflection/bremsstrahlung/NaivBremsstrahlung.cxx
```diff
@@ -1,11 +1,22 @@
 #include "PROPOSAL/scattering/stochastic_deflection/bremsstrahlung/NaivBremsstrahlung.h"
 #include "PROPOSAL/Constants.h"
-#include <iostream>
+
 using namespace PROPOSAL;
 
 std::array<double, 2>
```
maxnoe, line 6: `std::array` seems like a strange choice here. C++ normally encourages `std::pair` or `std::tuple` with `std::make_pair` and `std::make_tuple` for multiple return values.

MaxSac: I don't see the point.
These is a homogeneous type containers which are faster than resolving the underlying data type. What advantage do I get by using a tuple?

maxnoe: If this is an API where the two values have specific meaning and should all be the same, using a struct would probably be the best option. 

```
struct Deflection {
    double find_a_good_name_for_this1;
    double find_a_good_name_for_this1;
}
```

maxnoe: Compiled code will basically be the same. So it's mostly a matter of test and consistency. I would favor the struct approach as you can give the fields a semantic meaning and you can document it using doxygen at the struct.

MaxSac: Doesn't that imply that every stl container is unnecessary?
Maybe an enum should be created which defines the access to the elements explicitly. Otherwise I don't see the need to create a separate object for it.

maxnoe: The struct approach will also give you better type safety, you cannot call a function that just happens to also return an array of two doubles in a context where the output of a deflection would be required.

pascalgutjahr:
I changed the names of parametrizations. I still use NaivIonization because it's just the naive calculation of 4 momentum conservation. 
Now I use the PROPOSAL constants and the return value of each parametrization is a struct with zenith and azimuth.

include/PROPOSAL/scattering/ScatteringMultiplier.h
```diff
@@ -7,12 +7,12 @@ class ScatteringMultiplier : public Scattering {
...
 
-    std::array<double, 2> _scale_deflect(std::array<double, 2>& angles, InteractionType t) override
+    DirectionChangeAngular _scale_deflect(DirectionChangeAngular& angles, InteractionType t) override
     {
         for (auto m : stochastic_deflect) {
             if (m.first == t) {
-                std::get<0>(angles) *= m.second;
-                assert(std::get<0>(angles) <= PI);
+                angles.zenith *= m.second;
+                assert(angles.zenith <= PI);
```
maxnoe, line 15: Why is the assert here? these are not evaluated in optimized compiled code.

pascalgutjahr: Actually I don't know why here is an assert because I haven't implemented it, but instead we can use: 
`if (angles.zenith >= PI) { angles.zenith = 2*PI - angles.zenith};`

maxnoe: zenith is not defined in [0, 2π], only in [0, π] and a zenith larger than π is an error (or if small, due to floating point precision).

pascalgutjahr: That's correct and therefore I suggested to clip the angle. To be completely correct here we should have to check if zenith is higher than PI and smaller than 2PI. 
It is not obvious what to do with scaled angles which are higher than PI because in the most cases the deflection angles are very small but there are a few deflections which can reach quite high angles. Maybe we can set scaled angles to PI if they are higher.

MaxSac: > Why is the assert here? these are not evaluated in optimized compiled code.

In general, a multiplier that produces angles greater than pi is too large. For studies, only multipliers that will not go outside the definition range should be used and this should be checked by the user before. The assert exists only to simplify the debugging. 

I think there is no possibility to trade deflection angles larger than 180 degrees. The consistent possibility is to throw an assert that the angle distribution is broken.

sudojan:
Can you add Unit Tests and Reproducibility Tests?

pascalgutjahr:
I created a kind of unit test/tutorial jupyter notebook for this deflection method. But it's not possible to upload it via the comment field. Shall I push it to this fork?

MaxSac:
Maybe you can export it to a python document and push this one? 
Something like 
`jupyter nbconvert --to python notebookname.ipynb`
should do the work for you.

MaxSac:
Is this still wip or should it already be pulled to avoid divergences.?

You could open an issue where it is pointed out that the tests still have to be poured into cpp.

maxnoe:
> but it's not possible to upload it via the comment field. Shall I push it to this fork?


you can zip it and upload the zip file

pascalgutjahr:
> Is this still wip or should it already be pulled to avoid divergences.?
This is the check/tutorial notebook which I converted to a python file

> you can zip it and upload the zip file
Here I also uploaded the notebook in a zip

[deflection_check.ipynb.zip](https://github.com/tudo-astroparticlephysics/PROPOSAL/files/6090979/deflection_check.ipynb.zip)

include/PROPOSAL/math/MathMethods.h
```diff
@@ -171,7 +172,13 @@ std::pair<double, double> welfords_online_algorithm(double& newValue, unsigned i
 /// @return         Sampled value
 
 double SampleFromGaussian(double mean, double sigma, double rnd,
-                          double min = -std::numeric_limits<double>::infinity(),
-                          double max = std::numeric_limits<double>::infinity());
+                          double min = -INF,
+                          double max = INF);
```
MaxSac, line 176: I think for `INF` the `<cmath>` is required.

src/PROPOSAL/scattering/stochastic_deflection/pairProd/KelnerPairProduction.cxx
```diff
@@ -0,0 +1,36 @@
+#include "PROPOSAL/scattering/stochastic_deflection/pairProd/KelnerPairProduction.h"
+#include "PROPOSAL/Constants.h"
+#include "math.h"
```
MaxSac, line 3: i think we should use `<cmath>` and `<random>` instead

MaxSac (changes requested):

include/PROPOSAL/math/MathMethods.h
maxnoe: Also what is wrong with using the modern c++ version?

Jean1995: INF actually is PROPOSAL::INF, where it is defined as `std::numeric_limits<double>::infinity()`

src/PROPOSAL/scattering/stochastic_deflection/pairProd/KelnerPairProduction.cxx
Jean1995: I forgot to change that here

include/PROPOSAL/math/MathMethods.h
MaxSac: sorry i had overlooked it and thought it was a c macro.

MaxSac (approved):

Issue #125, opened 2021-02-19
photohadronic loss model addition


ibsafa:
An updated parameterization of the F2 functions based on fits to HERA data was made available in 2013. This description fits recent data better, and would be a great addition to the available models in PROPOSAL. 

The parameterization is very similar to the ALLM97 model, the only changes being the values of the fit parameters, see https://arxiv.org/pdf/1302.6119.pdf Eq. 2 for the F2 function, and Table 1. for the fitted parameter values.

asandrock:
Hi Ibrahim, this parametrization is in fact rather different from ALLM97, which you can see for instance from the fact that this parametrization contains 7 parameters, while the ALLM parametrization has 23. Moreover, this parametrization diverges for Q^2 -> 0, which seems problematic for use in PROPOSAL, as the minimum value of Q^2 is small.

There actually has been an update on the ALLM parametrization in Abt et al. (2017) [https://doi.org/10.1103/PhysRevD.96.014001] which really only changes the values of the fit parameters (Table VIII). Would you be satisfied if we implemented this model instead, or are you specifically looking for a parametrization that saturates the Froissart bound?

ibsafa:
Hi Alex 
Sorry I should have been clearer. What I meant by similar is that the implementation of this model, in terms of integration limits etc. in the code, is the same, the only difference being the F2 function. So it’s just a matter of coding these parameters in and calling the new F2 function. I’ve implemented this model in MMC as described and it works well. I have also checked it against an independent analytic calculation of the losses according to this model.
As for the divergence, this model specifically deals with small x, large Q^2 where the ALLM parametrization diverges, so they are complementary. 

We are using this for extremely high energy taus, where losses become dominant and the photohadronic losses in particular take over. Using this model vs. ALLM shows a factor of ~2 change in the range at high energies so it’s important to have both of them when doing calculations in this energy range.

asandrock:
I calculated the average energy loss for muons and taus with this parametrization, with ALLM 97 and with the newer parametrizations by Abt et al. The average energy loss is absurdly high, as I expected for a parametrization that diverges for Q^2 -> 0. There is a newer paper by Block et al. from 2014 [http://dx.doi.org/10.1103/PhysRevD.89.094027], where they use a parametrization that smoothly connects to the photoabsorption cross section parametrization from Block & Halzen (2004). I have shown this parametrization below as BDH; I have taken only the asymptotic component into account and ignored the valence quarks.
[...]

PR #128, opened 2021-03-11
Implementation of new Photonuclear Parametrizations

asandrock:
New photonuclear parametrizations according to Block, Durand & Ha (2014) and Abt et al. (2017)

### DIFF
--- a/private/PROPOSAL/crossection/parametrization/PhotoQ2Integration.cxx
+++ b/private/PROPOSAL/crossection/parametrization/PhotoQ2Integration.cxx
@@ -140,6 +140,7 @@ Q2_PHOTO_PARAM_INTEGRAL_IMPL(AbramowiczLevinLevyMaor97)
 Q2_PHOTO_PARAM_INTEGRAL_IMPL(ButkevichMikhailov)
 Q2_PHOTO_PARAM_INTEGRAL_IMPL(RenoSarcevicSu)
 Q2_PHOTO_PARAM_INTEGRAL_IMPL(AbtFT)
+Q2_PHOTO_PARAM_INTEGRAL_IMPL(BlockDurandHa)
 
 // ------------------------------------------------------------------------- //
 // Abramowicz Levin Levy Maor 91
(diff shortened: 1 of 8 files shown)


private/PROPOSAL/crossection/parametrization/PhotoQ2Integration.cxx
```diff
@@ -616,4 +618,225 @@ double PhotoRenoSarcevicSu::FunctionToQ2Integral(double energy, double v, double
...
+    // differential cross section from Dutta et al.
+    // Phys. Rev. D 63 (2001), 094020
+    // eq. 3.11
+    // eq. 3.4
+    // \frac{d^2 \sigma}{dQ^2 dx} = \frac{4\pi\alpha^2}{Q^4} \frac{F_{2, Proton}}{v}
+    //     ( 1 - v + \frac{M x v}{2E}
+    //     - \frac{v^2}{2} (1 - \frac{2m_{particle}}{Q^2})
+    //          \frac{1 + \frac{4 M^2 x^2}{Q^2}}{1 + R})
+    double result = ME * RE / Q2;
+    result *= result * 4 * PI * structure_function_nucleus / v *
```
maxnoe, line 749: Is this really correct? `result *= result ... ` look looks suspicious.

sudojan: yes, that's quite often the way how things are squared in PROPOSAL.
It's probably not code-reader friendly

maxnoe: Maybe worth taking a look to implement a `pow(double, int)` ?

https://stackoverflow.com/a/13771544/3838691

maxnoe: Actually, all current compilers just replace `pow(x, 2)` with `x * x`: https://godbolt.org/z/a3c5sY

maxnoe: So the above is only beneficial for integer powers > 2

sudojan (approved):

sudojan:
This will close issue #125

PR #147, opened 2021-04-19
:bug: nlohmann_json deleted a optional flag in conan. This caused a u…

MaxSac:
…ndefinded var error fix by this commit.


Jean1995 (approved):

Jean1995:
This means we should be able to use the json_fwd.hpp again. I'll try to do that asap.

PR #149, opened 2021-04-20
Use json_fwd.hpp for forward declarations whenever possible

Jean1995:
Now that [this PR in conan-center-index](https://github.com/conan-io/conan-center-index/pull/5188) has been approved, we can use `json_fwd.hpp` in our code when we only need the forward declarations from nlohmann_json (for example in `.h` files). I've been going through the code to find as many files as possible where we can use `json_fwd.hpp` instead of the full `json.hpp` to save compilation time. I had to add some headers from the std library that were missing now.

### DIFF
--- a/src/PROPOSAL/PROPOSAL/geometry/Box.h
+++ b/src/PROPOSAL/PROPOSAL/geometry/Box.h
@@ -30,7 +30,7 @@
 #pragma once
 
 #include "PROPOSAL/geometry/Geometry.h"
-#include <nlohmann/json.hpp>
+#include <nlohmann/json_fwd.hpp>
 
 
 namespace PROPOSAL {
--- a/src/PROPOSAL/PROPOSAL/geometry/Box.h
+++ b/src/PROPOSAL/PROPOSAL/geometry/Box.h
@@ -30,8 +30,6 @@
 #pragma once
 
 #include "PROPOSAL/geometry/Geometry.h"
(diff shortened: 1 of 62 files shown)


src/PROPOSAL/PROPOSAL/geometry/Box.h
```diff
@@ -30,7 +30,7 @@
 #pragma once
 
 #include "PROPOSAL/geometry/Geometry.h"
-#include <nlohmann/json.hpp>
+#include <nlohmann/json_fwd.hpp>
```
sudojan, line 33: `json_fwd.hpp` is directly included here, but not in Cylinder or Sphere,
not even in Geometry.h
one could make this more consistent

MaxSac:
Would it be possible to compare the times it takes to build 

a) the library 
b) an example 

before and after this PR and share the results here? I would be very interested.

Jean1995:
> Would it be possible to compare the times it takes to build
> 
> a) the library
> b) an example
> 
> before and after this PR and share the results here? I would be very interested.

Building PROPOSAL without tests and without python bindings:

**master:** 2.50 min (real)
**json_fwd:** 2.11 min (real)

Building PROPOSAL with tests and without python bindings:

**master:** 5.09 min (real)
**json_fwd:** 4.19 min (real)

MaxSac:
thank you for the work, it is really interesting :+1:

MaxSac (approved):

PR #151, opened 2021-04-22
Update water/ice properties and add liquid argon.

austinschneider:
Since PROPOSAL's original implementation the properties of liquid water and ice have been updated in the PDG. This pull request updates these quantities with those found here: https://pdg.lbl.gov/2020/AtomicNuclearProperties/HTML/water_liquid.html
This pull request also adds liquid argon as a medium so that PROPOSAL may be used by experiments like DUNE.

### DIFF
--- a/src/PROPOSAL/PROPOSAL/medium/Medium.h
+++ b/src/PROPOSAL/PROPOSAL/medium/Medium.h
@@ -143,6 +143,7 @@ MEDIUM_DEF(Copper)
 MEDIUM_DEF(Uranium)
 MEDIUM_DEF(Paraffin)
 MEDIUM_DEF(Air)
+MEDIUM_DEF(LiquidArgon)
 
 // #<{(|
 // * initialize ANTARES water
@@ -197,6 +198,7 @@ enum Medium_Type {
     PARAFFIN,
     ANTARESWATER,
     CASCADIABASINWATER,
+    LIQUIDARGON,
 };
 } // namespace PROPOSAL
 
@@ -216,6 +218,7 @@ static const std::array<std::string, 14> Medium_Name = {
     "paraffin",
     "antareswater",
     "cascadiabasinwater",
+    "liquidargon",
 };
 } // namespace PROPOSAL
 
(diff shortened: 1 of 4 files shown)


austinschneider:
I don't quite understand the build error from travis-ci with this latest commit a2c4e2b. Perhaps someone else can take a look.

Jean1995:
> I don't quite understand the build error from travis-ci with this latest commit [a2c4e2b](https://github.com/tudo-astroparticlephysics/PROPOSAL/commit/a2c4e2bc71931e5acaac9c44da4f4a4911c0539d). Perhaps someone else can take a look.

Our local server infrastructure, where the TestFiles are hosted, is currently down for maintenance. Maybe travis is killing the jobs after too many timeouts. I will restart the build tomorrow when the services are back online.

austinschneider:
> Our local server infrastructure, where the TestFiles are hosted, is currently down for maintenance. Maybe travis is killing the jobs after too many timeouts. I will restart the build tomorrow when the services are back online.

Thanks @Jean1995

maxnoe:
The dataserver is up and running again.

maxnoe:
Tests are running but failing. I assume this is because the test data that is used for regression tests.

This leads me to a general comment: I think it is not good to change the definition of existing media. Please create a new medium like `WaterPDG<year>`. Otherwise people might be confused by subtle changes when updating.

sudojan:
> Tests are running but failing. I assume this is because the test data that is used for regression tests used the old medium definitions.
> 
> This leads me to a general comment: I think it is not good to change the definition of existing media. Please create a new medium like `WaterPDG<year>`. Otherwise people might be confused by subtle changes when updating.
> 
> // same for ice

The old medium definitions were also taken from the PDG, but the PDG has updated some values since the MMC time at the beginning of 2000. I would not create even more media of the same type each time the PDG makes a new update changing slightly some constants. If users would like to stick to the old constants, they can checkout the old version.
We should change the cross sections, since this is our core and there, people can choose between different calculations for the same thing for their purpose. But the medium can be updated.
At least, this is my opinion.

Regarding the tests, one can produce a new set of TestFiles with the updated medium constants. We are already in a discussion to revise these ci tests.

sudojan:
You could mention the new medium LiquidArgon also in `docs/config_docu.md`

maxnoe:
>  I would not create even more media of the same type each time the PDG makes a new update changing slightly some constants. 

This is exactly what I'd do and what most scientific packages do. See e.g. the astropy constants module that allows choosing the CODATA year (and not forces people to use the current version of CODATA with a current version of astropy):

https://docs.astropy.org/en/stable/constants/

Also, it should be clear where values come from. So adding doxygen docs with a reference for examlpe.

Jean1995:
How about we have a `WaterPDG20XX` object (with the correct year) for the old water implementation, a `WaterPDG2020` for the new water implementation, and furthermore define `Water` as a alias for the most recent water implementation. This way, users can decide whether they want to use a specific version of the medium (that will not change in future releases) or just always the most current medium constants.

Issue #143, opened 2021-04-15
CI Tests on MacOS


sudojan:
There are some/many users, who are using PROPOSAL on a MacOS.
Therefore a build on MacOS should be included in the CI.
Some MacOS related issues could have also been identified earlier.
Shouldn't be too difficult adding this on travis.

maxnoe:
It isn't difficult,  but expensive on Travis.

As discussed some time ago with @MaxSac , due to the pricing changes on Travis, you should switch to using github actions instead, which also offers macos builds.

Jean1995:
Maybe we can also provide a Windows build there, which is not that well supported by Travis at the moment.

PR #155, opened 2021-05-03
Setup GitHub actions

MaxSac:


Jean1995 (changes requested):
I believe we do not test the python interface at the moment, we are not even building it, right? We should change that before merging.

.github/workflows/pypi_depoloy.yml
```diff
@@ -0,0 +1,28 @@
...
+on:
+  page_build:
+  release:
+    types:
+      - created
+
+jobs:
+
+    build-n-publish:
+      name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
```
Jean1995, line 12: we do not distribute to TestPyPI anymore

.github/workflows/pypi_depoloy.yml
```diff
@@ -0,0 +1,28 @@
...
+      - uses: actions/checkout@master
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+      - name: Install build
+        run: python -m pip install build
+      - name: Build source tarball
+        run: python -m build -s
+      - name: Publish distribution 📦 to Test PyPI
```
Jean1995, line 24: see above

MaxSac:
Closes #143

Jean1995 (approved):

PR #156, opened 2021-05-03
Fix travis CI for clang

Jean1995:
Both clang-6 and clang-10 have recently been broken on travis.

I fixed it according to the instructions from [this post](https://travis-ci.community/t/clang-10-was-recently-broken-on-linux-unmet-dependencies-for-clang-10-clang-tidy-10-valgrind/11527) in the travis forum.

To be honest, I am not quite happy that we need to make such a fix. So this would be another reason to change to github actions, as already suggested in issue #143. Maximilian already started some tests to make this work in MR #155.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,13 +64,9 @@ jobs:
       dist: focal
       name: Linux clang 6.0
       python: 3.9
-      addons:
-        apt:
-          sources:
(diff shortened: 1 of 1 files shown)


MaxSac:
This is now obsolete and we can close it, right? @Jean1995

Issue #48, opened 2020-02-21
Make Interpolation and Integration routines polymorph


Jean1995:
The current implementation of the Interpolation and Integrations routines are very static and could be vastly improved by rewriting them in a polymorphic way. This could not only improve the runtime but especially improve the readability of the code.

sudojan:
In MR #137 the interpolation is improved using the [CubicInterpolation library](https://github.com/MaxSac/cubic_interpolation)

PR #170, opened 2021-05-10
Change condition in Trapezoid3S from `<` to `<=`.

Jean1995:
This will avoid some Arithmetic Exceptions in some very rare cases (calling IntegrateWithRandomRatio on a function that is zero). This is probably not the perfect fix, but it helped in my case (and should not break any normal calculations)


src/PROPOSAL/detail/PROPOSAL/math/Integral.cxx
```diff
@@ -746,7 +746,7 @@ double Integral::Trapezoid3S(int n, double oldSum, int stepNumber)
                         {
                             determinant = std::sqrt(determinant);
                             approX      = (bEq + determinant) / aEq;
-
+                            // Why are both following conditions equal?
```
MaxSac, line 749: Maybe add a TODO so that the developer has a gerater change to find this place?
```suggestion
                            // TODO: Why are both following conditions equal?
```

MaxSac (changes requested):

MaxSac (approved):

sudojan:
Can you point out when this happens or provide an example script reproducing this issue`?

Jean1995:
> Can you point out when this happens or provide an example script reproducing this issue`?

```
#include <xmmintrin.h>
using namespace PROPOSAL;

double func_to_integral(double x) {
    return 0;
}

int main(){
    _MM_SET_EXCEPTION_MASK(_MM_GET_EXCEPTION_MASK() & ~_MM_MASK_INVALID); // enable/disable floating point exceptions
    Integral integral;
    auto val = integral.IntegrateWithRandomRatio(0, 10, func_to_integral, 3, 0.5);
    std::cout << val << std::endl;`
}
```

This is the example to use on clang. Commenting/uncommenting the line with the EXCEPTION_MASK causes PROPOSAL to throw an `EXC_ARITHMETIC`.

However, I just realized that the fix in this PR is not sufficient (it code will make another division by zero in another line). So I have to investigate this a little bit further to solve the issue....

MaxSac (approved):

sudojan:
> 
> 
> The new fix is probably more intuitive. To _really_ fix the problem behind this, one would probably have to start digging into this (17 years old) code much deeper...

This is probably related to #48 and a complete overhall of the integration process. Maybe they can be outsourced like the interpolation, maybe even inside the CubicInterpolation library.
But for now it seems to be sufficient.

Have you also checked if this change in the integration routine does also pass the test of the cross sections, that are excluded for the continuous integration?

Jean1995:
> > The new fix is probably more intuitive. To _really_ fix the problem behind this, one would probably have to start digging into this (17 years old) code much deeper...
> 
> This is probably related to #48 and a complete overhall of the integration process. Maybe they can be outsourced like the interpolation, maybe even inside the CubicInterpolation library.
> But for now it seems to be sufficient.
> 
> Have you also checked if this change in the integration routine does also pass the test of the cross sections, that are excluded for the continuous integration?

Not yet, but I will try to let these test run today and keep you updated.

Jean1995:
As I expected, I can confirm that at least all the Integral_Tests from the remaining cross sections (Bremsstrahlung, Epairproduction, Mupairproduction, Photonuclear) are not failing as well with these changes.

sudojan (approved):

PR #173, opened 2021-05-20
add pybindings for the `DirectionChangeAngular`

sudojan:
To call the stochastic deflection in python the return type was not pybinded resulting in:
`TypeError: Unable to convert function return value to a Python type! The signature was
	(self: proposal.scattering.StochasticDeflection, initial_energy: float, loss_energy: float, random_numbers: List[float]) -> PROPOSAL::DirectionChangeAngular`

an example code would be
```
import proposal as pp

mudef = pp.particle.MuMinusDef()
med = pp.medium.StandardRock()
param = pp.make_stochastic_deflection('kelnerpairproduction', mudef, med)

 deflect = param.stochastic_deflection(1e6, 1e5, [0.1, 0.2, 0.3, 0.1])
```

### DIFF
--- a/src/PROPOSAL/PROPOSAL/math/Spherical3D.h
+++ b/src/PROPOSAL/PROPOSAL/math/Spherical3D.h
@@ -34,4 +34,22 @@ namespace PROPOSAL  {
     protected:
         void print(std::ostream&) const override;
      };
+
+
+    /*!
+     * Container for a spherical vector with radius 1
+     *
+     * \param zenith
+     * \param azimuth
+     */
+     struct UnitSphericalVector{
+        double zenith;
+        double azimuth;
+    
+        constexpr UnitSphericalVector(const UnitSphericalVector&) = default;
+        constexpr UnitSphericalVector(double zenith, double azimuth)
+            : zenith(zenith), azimuth(azimuth) {};
+        constexpr UnitSphericalVector(const Spherical3D& s)
+            : zenith(s.GetZenith()), azimuth(s.GetAzimuth()) {};
+     };
 } // namespace PROPOSAL
--- a/src/PROPOSAL/PROPOSAL/math/Spherical3D.h
+++ b/src/PROPOSAL/PROPOSAL/math/Spherical3D.h
@@ -43,10 +43,10 @@ namespace PROPOSAL  {
      * \param azimuth
      */
      struct UnitSphericalVector{
-        double zenith;
-        double azimuth;
+        double zenith = 0;
+        double azimuth = 0;
     
-        UnitSphericalVector() = default;
+        constexpr UnitSphericalVector() = default;
         constexpr UnitSphericalVector(double zenith, double azimuth)
             : zenith(zenith), azimuth(azimuth) {};
         constexpr UnitSphericalVector(const Spherical3D& s)
--- a/src/PROPOSAL/PROPOSAL/math/Spherical3D.h
+++ b/src/PROPOSAL/PROPOSAL/math/Spherical3D.h
@@ -15,8 +15,8 @@ namespace PROPOSAL  {
 
 
         auto GetRadius() const {return coordinates[Radius];}
-        auto GetAzimuth() const {return coordinates[Azimuth];}
-        auto GetZenith() const {return coordinates[Zenith];}
+        constexpr auto GetAzimuth() const {return coordinates[Azimuth];}
+        constexpr auto GetZenith() const {return coordinates[Zenith];}
         void SetRadius(double radius) {coordinates[Radius] = radius;}
         void SetAzimuth(double azimuth) {coordinates[Azimuth] = azimuth;}
         void SetZenith(double zenith) {coordinates[Zenith] = zenith;}
@@ -46,7 +46,7 @@ namespace PROPOSAL  {
         double zenith;
         double azimuth;
     
-        constexpr UnitSphericalVector(const UnitSphericalVector&) = default;
+        UnitSphericalVector() = default;
         constexpr UnitSphericalVector(double zenith, double azimuth)
             : zenith(zenith), azimuth(azimuth) {};
         constexpr UnitSphericalVector(const Spherical3D& s)
--- a/src/PROPOSAL/detail/PROPOSAL/scattering/stochastic_deflection/bremsstrahlung/TsaiApproximationBremsstrahlung.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/scattering/stochastic_deflection/bremsstrahlung/TsaiApproximationBremsstrahlung.cxx
@@ -3,7 +3,7 @@
 #include <cmath>
 using namespace PROPOSAL;
 
-DirectionChangeAngular 
+UnitSphericalVector 
 stochastic_deflection::TsaiApproximationBremsstrahlung::CalculateStochasticDeflection(
     double e_i, double e_f, std::vector<double> const& rnd) const
 {
@@ -15,5 +15,5 @@ stochastic_deflection::TsaiApproximationBremsstrahlung::CalculateStochasticDefle
     auto theta_photon = mass / e_i * r;
     auto theta_muon = epsilon / e_f * theta_photon;
 
-    return  DirectionChangeAngular {theta_muon, 2 * PI * rnd[1]};
+    return  UnitSphericalVector {theta_muon, 2 * PI * rnd[1]};
 }
--- a/src/PROPOSAL/PROPOSAL/PROPOSAL.h
+++ b/src/PROPOSAL/PROPOSAL/PROPOSAL.h
@@ -156,7 +156,6 @@
 #include "PROPOSAL/scattering/ScatteringFactory.h"
 #include "PROPOSAL/scattering/ScatteringMultiplier.h"
 #include "PROPOSAL/scattering/ScatteringFactory.h"
-#include "PROPOSAL/scattering/DirectionChange.h"
 
 #include "PROPOSAL/Constants.h"
 #include "PROPOSAL/EnergyCutSettings.h"
--- a/src/PROPOSAL/PROPOSAL/scattering/DirectionChange.h
+++ b/src/PROPOSAL/PROPOSAL/scattering/DirectionChange.h
@@ -1,8 +1,0 @@
-#pragma once 
-
-namespace PROPOSAL {
-struct DirectionChangeAngular {
-    double zenith;
-    double azimuth;
-};
-}
--- a/src/PROPOSAL/PROPOSAL/scattering/Scattering.h
+++ b/src/PROPOSAL/PROPOSAL/scattering/Scattering.h
@@ -53,8 +53,8 @@ class Scattering {
     }
 
 protected:
-    virtual DirectionChangeAngular _scale_deflect(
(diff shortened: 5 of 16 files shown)


sudojan:
btw, why does the stochastic deflection requires a medium?
For all 4 parametrizations implemented so far, none of them is using the medium.

sudojan:
another question: why is the return type not a Spherical Vector?
why defining an extra class just as a container for zenith and azimuth?
Well the radius is not used, here but it would be more intuitive to return a spherical coordinate, the rest is not too much of an overhead?

maxnoe:
Astropy has two classes for this, SphericalRepresentation and UnitSphericalRepresentation. Maybe that's also a good choice here? Adding a UnitSphericalVector for this?

MaxSac:
> btw, why does the stochastic deflection requires a medium?
> For all 4 parametrizations implemented so far, none of them is using the medium.

Templates like 
```
template <typename ...Args>
auto make_deflection(std::string const &name, Args... args) {
...
};
```
can't be pybindet. Therefore, we have to make a guess what the parameters for a good factory function are. 
I could imagine that there might be some deflection functions that depend on the atomic number of the component. 
@sudojan In general, the function can be overloaded if it is reasonably defined.

MaxSac:
> another question: why is the return type not a Spherical Vector?
> why defining an extra class just as a container for zenith and azimuth?
> Well the radius is not used, here but it would be more intuitive to return a spherical coordinate, the rest is not too much of an overhead?

The spherical vector was introduced after the then MR for deflection. @Jean1995  I think we can improve this, right?

MaxSac:
> Astropy has two classes for this, SphericalRepresentation and UnitSphericalRepresentation. Maybe that's also a good choice here? Adding a UnitSphericalVector for this?

This would certainly be the best solution :wink:

MaxSac (changes requested):
- [x] unit spherical vector as return type
- [ ] better docstring in the pybindet `stochastic_deflection`

Jean1995:
> another question: why is the return type not a Spherical Vector?
> why defining an extra class just as a container for zenith and azimuth?
> Well the radius is not used, here but it would be more intuitive to return a spherical coordinate, the rest is not too much of an overhead?

Although I see you point, I am wondering: Could this return type be confusing for people? When you get a vector, you might believe that you get the new direction instead of just the deflection angles that you need to apply to the old direction.

Jean1995:
>     * [ ]  unit spherical vector as return type
> 
>     * [ ]  better docstring in the pybindet `stochastic_deflection`

Implementing a UnitSphericalVector shouldn't be too big of a problem, but we would need to revise the class structure at this moment. For example, `Vector3D` (the virtual base class) has a virtual function `void SetCoordinates(std::array<double, 3> val) {coordinates = val;}`, which would make no sense anymore for a UnitSphericalVector. The solution will probably by just to remove these types of virtual functions.

maxnoe:
I don't think having a vector implemented having a `std::array<double, 3>` is the best option. Why not use a `struct {x, y, z}`, `struct {r, theta, phi}`, `struct {theta, phi}` ?

Jean1995:
> I don't think having a vector implemented having a `std::array<double, 3>` is the best option. Why not use a `struct {x, y, z}`, `struct {r, theta, phi}`, `struct {theta, phi}` ?

Yes, also having individual Getter and Setter methods with these structs as return types for each Vector type is probably more intuitive than an arbitrary std::array

maxnoe:
Don't use getters and setters. Just use a struct with public data members for such simple data structures. It makes things a whole lot easier (and more efficient, since you copy less)

maxnoe:
Something like this: https://godbolt.org/z/nrWPner9M

src/PROPOSAL/PROPOSAL/math/Spherical3D.h
```diff
@@ -34,4 +34,22 @@ namespace PROPOSAL  {
...
+     * Container for a spherical vector with radius 1
+     *
+     * \param zenith
+     * \param azimuth
+     */
+     struct UnitSphericalVector{
+        double zenith;
+        double azimuth;
+    
+        constexpr UnitSphericalVector(const UnitSphericalVector&) = default;
```
MaxSac, line 49: This is already the implicitly-declared copy constructor and we can spare this line.
I think it's no good idea to explicit write it down because it's seems that it is necessary for each class.

MaxSac (changes requested):
- [x] delete self defined copy cstr which is a implicitly-declared copy constructor

src/PROPOSAL/PROPOSAL/math/Spherical3D.h
maxnoe: right, sorry, I forgot that the copy constructor is not removed when you define alternative constructors.

maxnoe: But the default constructor has to be implemented back.

src/PROPOSAL/PROPOSAL/math/Spherical3D.h
```diff
@@ -34,4 +34,22 @@ namespace PROPOSAL  {
...
+        constexpr UnitSphericalVector(double zenith, double azimuth)
+            : zenith(zenith), azimuth(azimuth) {};
+        constexpr UnitSphericalVector(const Spherical3D& s)
+            : zenith(s.GetZenith()), azimuth(s.GetAzimuth()) {};
```
MaxSac, line 53: For this constexpr copy cstr, the getter functions must also be constexpr functions.

MaxSac (changes requested):

src/PROPOSAL/PROPOSAL/math/Spherical3D.h
sudojan: well, it worked with Apple Clang 10

sudojan:
```
134: C++ exception with description "input stream error" thrown in the test body.
134: [  FAILED  ] Ionization.Test_of_dEdx_Interpolant (12632 ms)
```
I doubt, this has to do with my changes

Jean1995:
> ```
> 134: unknown file: Failure
> 134: C++ exception with description "input stream error" thrown in the test body.
> 134: [  FAILED  ] Ionization.Test_of_dEdx_Interpolant (12632 ms)
> ```
> 
> I doubt, this has to do with my changes

Github actions sometimes gets a hiccup with this UnitTest, I do not understand why since I can't reproduce it, but restarting the tests normally helps

MaxSac (approved):

src/PROPOSAL/PROPOSAL/math/Spherical3D.h
```diff
@@ -34,4 +34,22 @@ namespace PROPOSAL  {
...
+     * Container for a spherical vector with radius 1
+     *
+     * \param zenith
+     * \param azimuth
+     */
+     struct UnitSphericalVector{
+        double zenith;
+        double azimuth;
+    
+        UnitSphericalVector() = default;
```
maxnoe, line 49: shouldn't this also be constexpr?

src/PROPOSAL/detail/PROPOSAL/scattering/stochastic_deflection/bremsstrahlung/TsaiApproximationBremsstrahlung.cxx
```diff
@@ -15,5 +15,5 @@ stochastic_deflection::TsaiApproximationBremsstrahlung::CalculateStochasticDefle
     auto theta_photon = mass / e_i * r;
     auto theta_muon = epsilon / e_f * theta_photon;
 
-    return  DirectionChangeAngular {theta_muon, 2 * PI * rnd[1]};
+    return  UnitSphericalVector {theta_muon, 2 * PI * rnd[1]};
```
Jean1995, line 18:
```suggestion
    return UnitSphericalVector {theta_muon, 2 * PI * rnd[1]};
```

Jean1995 (changes requested):

Jean1995 (changes requested):
Now that the return type of `CalculateStochasticDeflection` is called `UnitSphericalVector` instead of `DirectionChangeAngular`, it should be made very clear in the "documentation" that we are still returning a direction change and not a (normalized) direction in a three-dimensional room. 

Maybe we can add a descriptive comment at least [in the pybindings](https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/50ac95a4d2aa1ef63749254adb069a763926d4e2/src/pyPROPOSAL/detail/scattering.cxx#L89), maybe also in the cpp file.

src/PROPOSAL/PROPOSAL/math/Spherical3D.h
sudojan: hmm, I tried but then I get the error
```defaulted definition of default constructor is not constexpr```

sudojan: ah, they need default values
but should they have default values?

sudojan:
> Now that the return type of `CalculateStochasticDeflection` is called `UnitSphericalVector` instead of `DirectionChangeAngular`, it should be made very clear in the "documentation" that we are still returning a direction change and not a (normalized) direction in a three-dimensional room.
> 
> Maybe we can add a descriptive comment at least [in the pybindings](https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/50ac95a4d2aa1ef63749254adb069a763926d4e2/src/pyPROPOSAL/detail/scattering.cxx#L89), maybe also in the cpp file.

regarding the comment in the pydocs:
I would prefer the energy loss instead of the final energy of the muon, since all cross sections (i.e. our physical input) are taking the energy loss. But it is just a preference, I could also live how it is done, yet
But this discussion is probably too forky from the main point of this MR.

sudojan:
> I don't think having a vector implemented having a `std::array<double, 3>` is the best option. Why not use a `struct {x, y, z}`, `struct {r, theta, phi}`, `struct {theta, phi}` ?

> Don't use getters and setters. Just use a struct with public data members for such simple data structures. It makes things a whole lot easier (and more efficient, since you copy less)

Maybe out of the Scope of this MR, but regarding these comments, should the Vector3D with Cartesian3D and Spherical3D also be structs with less overhead, e.g. Getters and Setters?

Jean1995:
> regarding the comment in the pydocs:
> I would prefer the energy loss instead of the final energy of the muon, since all cross sections (i.e. our physical input) are taking the energy loss. But it is just a preference, I could also live how it is done, yet
> But this discussion is probably too forky from the main point of this MR.

I see your point, but this would pretty much break the versioning. So we would probably need to safe this change for a major release, don't we?

Jean1995:
> Maybe out of the Scope of this MR, but regarding these comments, should the Vector3D with Cartesian3D and Spherical3D also be structs with less overhead, e.g. Getters and Setters?

If we use the dedicated struct for each class (instead of the `std::array<double, 3>), we could and should get rid of the `SetCoordinates()` function and the dedicated `Get...()` and `Set...()` function. However, we should still keep the functions such as `GetCartesianCoordinates()` to provide coordinate transformations. 
But the same argument about the versioning applies here as well.

Jean1995 (approved):

PR #179, opened 2021-06-02
fix secondaries calc for photon and add theta calc to public funcions

MaxSac:


examples/Interaction.ipynb
```diff
@@ -329,7 +329,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.2"
```
Jean1995, line 332: It this a remnant?

MaxSac: yes it's a remnant :sweat_smile:

Jean1995 (approved):

sudojan:
Maybe this should also be described in the [docs](https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/master/docs/config_docu.md)

Jean1995:
> Maybe this should also be described in the [docs](https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/master/docs/config_docu.md)

What exactly do you mean should be described in the config_docu? I think there is no json configuration so far, is it?

PR #185, opened 2021-07-13
Fix boost interpolation error

Jean1995:
We noticed that the Propagation can still produce boost exceptions when executing the `cubic_splines::find_parameter` method, although only in very rare cases.

After some investigation, I found out the reason for this.
When trying to run `UtilityInterpolant::GetUpperLimit` for interpolated functions, the Newton Raphson method can run into local minima when the spline is overshooting, see for example here:

image
This happens only rarely when the sampled random number is very specific.

My first idea and attempt at a solution was to fall back to the integrated function, `UtilityIntegral`, however mixing Interpolation and Integration can lead to inconsistencies in the long run.
Instead, I implemented a fall back to a simple Bisection method. This will be slower than the NewtonRaphson method, but more stable. To avoid that this fallback happens too often and without knowledge of the user, a warning will be issued by the Logger.
I also added a UnitTest to test the implemented lines, showing that the method indeed works very nicely.

### DIFF
--- a/src/PROPOSAL/detail/PROPOSAL/propagation_utility/PropagationUtilityInterpolant.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/propagation_utility/PropagationUtilityInterpolant.cxx
@@ -99,7 +99,7 @@ double UtilityInterpolant::GetUpperLimit(double upper_limit, double rnd)
     try {
         return cubic_splines::find_parameter(
                 *interpolant_, integrated_to_upper - rnd, initial_guess);
-    } catch (...) {
+    } catch (std::runtime_error&) {
         Logging::Get("proposal.UtilityInterpolant")->warn(
                 "Newton-Raphson iteration in UtilityInterpolant::GetUpperLimit "
                 "failed. Try solving using bisection method.");
--- a/src/PROPOSAL/detail/PROPOSAL/propagation_utility/PropagationUtilityInterpolant.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/propagation_utility/PropagationUtilityInterpolant.cxx
@@ -8,6 +8,8 @@
 #include "PROPOSAL/Constants.h"
 #include "PROPOSAL/propagation_utility/PropagationUtilityInterpolant.h"
 #include "PROPOSAL/methods.h"
+#include "PROPOSAL/Logging.h"
+#include "PROPOSAL/math/MathMethods.h"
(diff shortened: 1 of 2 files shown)


src/PROPOSAL/detail/PROPOSAL/propagation_utility/PropagationUtilityInterpolant.cxx
```diff
@@ -94,8 +96,18 @@ double UtilityInterpolant::GetUpperLimit(double upper_limit, double rnd)
     initial_guess.x = NAN;
     initial_guess.upper = upper_limit;
 
-    return cubic_splines::find_parameter(
-        *interpolant_, integrated_to_upper - rnd, initial_guess);
+    try {
+        return cubic_splines::find_parameter(
+                *interpolant_, integrated_to_upper - rnd, initial_guess);
+    } catch (...) {
```
sudojan, line 102: This shouldn't be activated for any exception.

sudojan:
Maybe a more general question, shouldn't this be fixed in the CubicSpline Interpolation library?

Jean1995:
> Maybe a more general question, shouldn't this be fixed in the CubicSpline Interpolation library?

This is a good question. The answer is "probably yes". But in the end, this is not my decision, but the decision of @MaxSac. I could also make a pull request directly for the `cubic_interpolation` library where I fix this issue there 🙂 

The only disadvantage I see at the moment is that this is probably going to make logging harder or impossible, since there is currently no logging provided by the `cubic_interpolation` library (as far as I know). However, I think it should not be unnoticed for the user if the newton raphson method fails.

src/PROPOSAL/detail/PROPOSAL/propagation_utility/PropagationUtilityInterpolant.cxx
Jean1995: Is it possible to catch a specific boost exception in PROPOSAL if PROPOSAL does not directly depend on boost?
I agree with your point, but I am not sure if this is possible.

sudojan (approved):

sudojan:
So, this is then finally the preferred approach, since #187 does not solve the problem and #188 takes much longer.
Is this correct?

Jean1995:
> So, this is then finally the preferred approach, since #187 does not solve the problem and #188 takes much longer.
> Is this correct?

Yes

PR #187, opened 2021-07-20
Alternative to #185 modify the lower lim of the interaction obj.

MaxSac:


src/PROPOSAL/PROPOSAL/crosssection/CrossSection.h
```diff
@@ -228,9 +229,20 @@ class CrossSection : public CrossSectionBase {
...
     double lower_energy_lim;
+    double min_stochastic_energy;
     InteractionType interaction_type;
     std::string param_name;
 
+private:
+    template <typename T> static double CalculateMinStochasticEnergy(T const& m)
+    {
+        double min_energy = std::numeric_limits<double>::infinity();
+        for (auto const& i : m)
```
maxnoe, line 240: This could just be std min_element

maxnoe:
```
    return std::min_element(
        m.cbegin(), m.cend(),
        [](const auto& kv1, const auto& kv2){return kv1.second.energy <= kv2.second.energy;}
    )->second.energy;
```

src/PROPOSAL/PROPOSAL/crosssection/CrossSection.h
```diff
@@ -228,9 +229,20 @@ class CrossSection : public CrossSectionBase {
     std::unique_ptr<std::vector<std::tuple<double, de2dx_ptr>>> de2dx;
 
     double lower_energy_lim;
+    double min_stochastic_energy;
     InteractionType interaction_type;
     std::string param_name;
 
+private:
+    template <typename T> static double CalculateMinStochasticEnergy(T const& m)
```
Jean1995, line 237: Does this need to be a template? Isn't T always the same type?

src/PROPOSAL/PROPOSAL/crosssection/CrossSectionVector.h
```diff
@@ -13,29 +13,42 @@ struct CrossSectionVector {
...
             });
-        return (*result)->GetLowerEnergyLim();
+        return (*r)->GetLowerEnergyLim();
+    }
+
+    template <typename CrossVec>
+    static double GetMinStochasticEnergy(CrossVec const& cross)
+    {
+        using val_t = typename std::decay<CrossVec>::type::value_type;
+        auto r = std::max_element(
```
Jean1995, line 39: Why is this the maximum instead of the minimum?

src/PROPOSAL/detail/PROPOSAL/propagation_utility/InteractionBuilder.cxx
```diff
@@ -1,40 +1,47 @@
 #include "PROPOSAL/propagation_utility/InteractionBuilder.h"
+#include "PROPOSAL/crosssection/CrossSection.h"
 #include "PROPOSAL/propagation_utility/DisplacementBuilder.h"
 #include "PROPOSAL/propagation_utility/PropagationUtilityInterpolant.h"
-#include "PROPOSAL/crosssection/CrossSection.h"
+#include <iostream>
```
Jean1995, line 5: relict?

src/PROPOSAL/PROPOSAL/propagation_utility/Interaction.h
```diff
@@ -47,6 +48,7 @@ class Interaction {
 
     double MeanFreePath(double energy);
 
+    auto GetLowerLim() const noexcept { return lower_lim; }
```
Jean1995, line 51: Calling this one `GetLowerLim` might be confusing since there already is a function with the same same in `CrossSectionVector` with a different meaning

src/PROPOSAL/PROPOSAL/crosssection/CrossSectionVector.h
Jean1995: Wait, this is `std::max_element`, but the comparison function object is defined the wrong way around, to it essentially is `std::min_element`, isn't it? 😅

src/PROPOSAL/PROPOSAL/crosssection/CrossSection.h
MaxSac: But this does not exist in c++14. It is a `c++17` feature :see_no_evil:

src/PROPOSAL/PROPOSAL/crosssection/CrossSectionVector.h
MaxSac: But this does not exist in c++14. It is a `c++17` feature :see_no_evil:

src/PROPOSAL/PROPOSAL/crosssection/CrossSection.h
MaxSac: Not really, but I was to lazy to search the corresponding type :sweat_smile:

src/PROPOSAL/PROPOSAL/crosssection/CrossSection.h
maxnoe: Ah, sorry, thought PROPOSAL was already C++17

src/PROPOSAL/PROPOSAL/crosssection/CrossSectionVector.h
maxnoe: But max_element as well. https://en.cppreference.com/w/cpp/algorithm/max_element

maxnoe: No, min_element exists also since basically forever: https://en.cppreference.com/w/cpp/algorithm/min_element

MaxSac: UFF, I have overseen the until. Good catch :+1:

src/PROPOSAL/PROPOSAL/crosssection/CrossSection.h
MaxSac: Done in 8b25d6febb5752eb552d19384c9e65bceade5f24.
But I am not sure if it's reduces the code complexity.

Jean1995:
Unfortunately, this does not solve the problem at hand.

This script, for example, is still failing for the same reason:

```
import proposal as pp
particle = pp.particle.MuMinusDef()
medium = pp.medium.Ice()
cuts = pp.EnergyCutSettings(np.inf, 1e-2, False)
cross = pp.crosssection.make_std_crosssection(particle, medium, cuts, True)
interaction = pp.make_interaction(cross, True)
E = 3180.6693833361928
rnd = 0.001549
interaction.energy_interaction(E, rnd)
```

The problem is that the introduced minimal stochastic energy is still too low to remove the overshooting of the interpolant. I am going to open another merge request with an alternative attempt to solve this problem.

Jean1995:
Solved by merging #185 into master

PR #188, opened 2021-07-26
Add fallback to UtilityIntegral::GetUpperLimit if UtilityInterpolant::GetUpperLimit fails

Jean1995:
After #187 did unfortunately not solve the original issue, and after the solution in #185 has not been completely satisfying, this is the third attempt to solve the problem with the boost error that is occasionally thrown when calling `cubic_splines::find_parameter` (see MR #185 for a more detailed description of the problem).

Here, I try to do a fallback to `UtilityIntegral::GetUpperLimit` in the case that `UtilityInterpolant::GetUpperLimit` fails.
To avoid numerical problems at the lowest energies, I had to introduce a cutoff if we get too close to the `lower_lim`.

A logging warning is thrown when this case occurs, however, it should only happen very rarely.


Jean1995:
Solved by merging #185 into master

PR #190, opened 2021-08-05
Repair weak interaction for propagation

Jean1995:
I figured out that when propagating a charged lepton, if it encounters a (charged) weak interaction, the propagation did not stop but the particle only lost a fraction of its energy. To fix this, the particle now loses all of its energy. The energy that goes into the neutrino and the hadronic cascade can now be calculated using the corresponding SecondariesCalculator.


sudojan:
Is the secondary caculator estimating the relative energy transfered to the neutrino or the hadronic cascade?

Jean1995:
> Is the secondary caculator estimating the relative energy transfered to the neutrino or the hadronic cascade?

It calculates the energy transferred to the hadronic cascade, see [here](https://github.com/tudo-astroparticlephysics/PROPOSAL/pull/190/files#diff-ce6c70c38cd6d0e6d012ee9c5c4b267eb5d96b11775bddf57ad9a20609fa52d8R40).
The rest of the energy (`1-v`) is assigned to the neutrino.

sudojan (approved):
Maybe it can be made more clear, that `CalculateRelativeLoss` corresponds to the energy transferes to the hadron (except of a comment in the source code), but the rest looks good.
Documentation is always an issue ;)

Issue #193, opened 2021-08-12
pip3 install fails on Ubuntu 20.04?


niess:
Hello,

first let me thank you for this great work and for making it available to the community :)

I had been using version 6.1.5 of PROPOSAL on Debian 9 without issues as far as I was concerned. Recently, I wanted to upgrade to the latest PyPI version of PROPOSAL, i.e. 7.0.5. But without success so far :(

Since Debian 9 seems too old w.r.t. several deps of PROPOSAL I ended up trying Ubuntu 20.04 first through docker and then through VirtualBox. In both cases I got the following issue. Installing with pip3 as:

```
pip3 install proposal
```
runs smoothly. However, importing the `proposal` Python package crashes as:

```Python
ImportError: $HOME/.local/lib/python3.8/site-packages/proposal.cpython-38-x86_64-linux-gnu.so: undefined symbol: _ZN13cubic_splines12CubicSplinesIdEC1ERKNS1_10DefinitionENSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESA_
```
Peeking at `proposal.cpython-38-x86_64-linux-gnu.so` with objdump indicates:
```
Dynamic Section:
  NEEDED               libpthread.so.0
  NEEDED               libstdc++.so.6
  NEEDED               libm.so.6
  NEEDED               libgcc_s.so.1
  NEEDED               libc.so.6
  NEEDED               ld-linux-x86-64.so.2
  NEEDED               libpython3.8.so.1.0
  SONAME               proposal.cpython-38-x86_64-linux-gnu.so
  RUNPATH              $ORIGIN
```
which does not exhibit any missing deps.

Note that with VirtualBox the Linux kernel is 5.11. Thus, this does not look like issue #186 . Python version is 3.8.10 and gcc version is 9.4 in case it could help?

Am I doing something wrong? What OS would you recommend to use?

Jean1995:
Hello! 

This sounds very much like a problem with the package manager conan. We have already experienced these problems on severral Linux systems before.

Could you try to run `conan profile update settings.compiler.libcxx=libstdc++11 default`, and then try to install proposal using `pip3 install proposal` again?

If this does not already help, there might be some cached files left that need to be rebuild. In this case, try to follow the steps that I suggest in issue #180.

I hope that these steps can already solve your issue!

niess:
@Jean1995 Thanks. 

[...]

Issue #180, opened 2021-06-09
Linker error when importing proposal after installation via pip on arch


Jean1995:
Doing `pip install proposal` on arch works, but importing proposal fails with the error

```
>>> import proposal as pp
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: /home/jean/.local/lib/python3.9/site-packages/proposal.cpython-39-x86_64-linux-gnu.so: undefined symbol: _ZN13cubic_splines12CubicSplinesIdEC1ERKNS1_10DefinitionENSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESA_
```

This has been brought to my attention by a user, and I've been able to reproduce it on my local arch.

Jean1995:
This issue has been caused by using the wrong conan settings. 
In the conan profile, `settings.compiler.libcxx` needs to be set to `libstdc++11`.

We were able to fix this issue by doing the following steps:

1. Uninstall old proposal version: `pip uninstall proposal`
2. Delete conan data folder: `rm -rf ~/.conan/data`
3. Update settings in conan profile (either manually or by doing `conan profile update settings.compiler.libcxx=libstdc++11 default`)
4. Install proposal via pip, but do not use cached version: `pip install proposal --no-cache-dir`
5. Reload proposal package, linking should work now.

PR #195, opened 2021-08-18
Improve setup.py

Jean1995:
This PR should fix two problems with the pypi installation:

1. If conan did not find the binaries for a certain configuration, the installation process failed since the `'--build=missing'` option has not been specified. This is really annoying for everyone trying to install proposal with pip on certain systems, for example on arch where very new packages are used.
2. We already discussed the problem a few times that per default, conan uses the old ABI while newer versions of gcc use the new ABI per default. If the user did not set up conan in advance, this caused linker errors when trying to import proposal, since the dependencies were built with the old ABI while PROPOSAL itself has been built with the new ABI. This PR detects whether the `compiler.libcxx` in conan is set to `libstdc++` (corresponding to the old ABI), and sets the according `CMAKE_CXX_FLAGS` so that PROPOSAL will also be built with the old ABI. If the user "correctly" set up conan in advance (or is on a different system, or is not using conan at all), nothing changes. This should fix common errors, leading to issues like #193 or #180.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,6 +35,19 @@ def get_cmake():
     raise OSError('You need cmake >= 3.9')
 
 
+def is_old_libcxx():
+    ''' if we are on gcc, we might be using an old library ABI '''
+    try:
+        compiler = sp.check_output(['conan', 'profile', 'get', 'settings.compiler', 'default'], encoding='UTF-8').rstrip('\n')
+        if compiler == 'gcc':
+            libcxx = sp.check_output(['conan', 'profile', 'get', 'settings.compiler.libcxx', 'default'], encoding='UTF-8').rstrip('\n')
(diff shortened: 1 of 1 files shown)


MaxSac:
This is a very nice PR and will improve the installation process :+1: 
There are a small little things I want to mention

- [ ] we are not checking if a minimal conan version is fullfilled
- [ ] we are not checking if a conan default profile exists

I will create a PR on top of this PR which will fix this issues :smile:

maxnoe:
Conan with version requirement is in pyproject.toml. No need to check in setup.py.

MaxSac:
Thx, I have missed that :sweat_smile:

MaxSac (approved):

PR #197, opened 2021-08-19
check if conan default profile is populated, otherwise populate it.

MaxSac:


Jean1995:
Thanks!

maxnoe:
Installing a python package should not have side effects like this. Is there no way to set this just for the current conan execution?

Jean1995:
> Installing a python package should not have side effects like this. Is there no way to set this just for the current conan execution?

What do you mean with "side effects"? Creating a (default) conan profile?

Instead of creating/using the default profile, we could create a profile with a specific name (e.g. "proposal_installation"), which is always used by pip and which the user can therefore also adapt (without interfering with the default profile). I am not sure whether it is more intuitive to use the default profile, or whether it is more intuitive NOT to use the default profile...

Looking at the conan documentation, conan should normally create the default profile by itself, quote from [here](https://docs.conan.io/en/latest/reference/config_files/default_profile.html):

> The initial values for these default settings are auto-detected the first time you run a conan command.

So maybe it is not even necessary that we create the default profile, since it should only be missing when a user explicitly deletes the default profile?

maxnoe:
Yes, creating anything in the user's home directory unrelated to the installation of the python package.

The whole point of pep 517 builds is the specification and isolation of the build. Creating configuration for the build system in the user's home directory is completely out of line.

Jean1995:
> Yes, creating anything in the user's home directory unrelated to the installation of the python package.
> 
> The whole point of pep 517 builds is the specification and isolation of the build. Creating configuration for the build system in the user's home directory is completely out of line.

In this case, the only possibility would be not to create the default profile, but to raise an error that there is no default profile.

maxnoe:
I don't understand. Even if there is a default profile, it is not okay to just change a user's system when they think they are just installing a python package. Persistently changing the user's conan profiles is not something a setup.py should do.

Jean1995:
But we do not change any settings? At the moment, if there are no default settings, conan will automatically detect them. Otherwise, we just read the profile values, but we do not make any changes.

maxnoe:
Sorry, misread. I thought you were still adding the libstdc++ setting to the profile.

PR #199, opened 2021-08-20
Keep initial lepton with updated energy instead of discarding particle

Jean1995:
We can not sample the secondary particles from photonuclear interactions yet, but we can at least keep the initial lepton with the updated energy.

### DIFF
--- a/src/PROPOSAL/PROPOSAL/secondaries/parametrization/photonuclear/Photonuclear.h
+++ b/src/PROPOSAL/PROPOSAL/secondaries/parametrization/photonuclear/Photonuclear.h
@@ -19,14 +19,22 @@ namespace secondaries {
             = PROPOSAL::InteractionType::Photonuclear;
         InteractionType GetInteractionType() const noexcept override { return type; }
 
-        std::vector<ParticleState> CalculateSecondaries(StochasticLoss, const Component&,
-                                                   std::vector<double>&) override
+        std::vector<ParticleState> CalculateSecondaries(
+                StochasticLoss loss, const Component&, std::vector<double>&) override
         {
             Logging::Get("proposal.Secondaries")
                     ->warn("PROPOSAL can not generate secondary particles"
                             "for a photonuclear interaction.");
             //TODO: Treatment for hadronic interactions
-            return std::vector<ParticleState>{};
+            auto primary_lepton = ParticleState();
+            primary_lepton.energy = loss.parent_particle_energy - loss.energy;
+            primary_lepton.type = primary_particle_type;
+            primary_lepton.time = loss.time;
+            primary_lepton.position = loss.position;
+            primary_lepton.direction = loss.direction;
+            primary_lepton.propagated_distance = 0.;
+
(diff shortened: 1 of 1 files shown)


src/PROPOSAL/PROPOSAL/secondaries/parametrization/photonuclear/Photonuclear.h
```diff
@@ -19,14 +19,22 @@ namespace secondaries {
...
 
-        std::vector<ParticleState> CalculateSecondaries(StochasticLoss, const Component&,
-                                                   std::vector<double>&) override
+        std::vector<ParticleState> CalculateSecondaries(
+                StochasticLoss loss, const Component&, std::vector<double>&) override
         {
             Logging::Get("proposal.Secondaries")
                     ->warn("PROPOSAL can not generate secondary particles"
                             "for a photonuclear interaction.");
             //TODO: Treatment for hadronic interactions
```
sudojan, line 28: if secondaries get returned, also the hadronic cascade should get returned; not just the lepton.
I am aware of the discussion in CORSIKA (see [here](https://gitlab.ikp.kit.edu/AirShowerPhysics/corsika/-/issues/411)), which can just handle real particles, but in my opinium the secondary output of PROPOSAl can also consist of hadronic pseudo particles containing the relevant information (Energy, position, direction, time and maybe also Q2).
Besides, it is not very likely, that one day PROPOSAl will calculate the real secondary particles of an hadronic cascade. Therefore, other calculation tools exist. The CORSIKA Problem should be dealt in CORSIKA.

The same of course also holds for the photo-hadronic, or weak interactions, or simply where a hadronic cascade is created.

Jean1995: Should PROPOSAL still issue a warning after we implemented the hadronic pseudoparticle / hadronic cascade into the Photonuclear secondaries calculator?

sudojan: They may be annoying, but I can also understand if they are left to signal the user, that not a real particle ws produced.

sudojan: Ah, and I see, that my other two concerns should be handled in another PR since
- for the weak interaction the secondary calculator is not yet fully implemented
- the photon-hadron interaction is not implemented at all
these are more relevat for a milestone 7.1.0

sudojan (approved):

src/PROPOSAL/PROPOSAL/particle/ParticleDef.h
```diff
@@ -103,6 +103,7 @@ enum class ParticleType : int {
     Monopole = 41,
     SMPPlus = 9500,
     SMPMinus = -9500,
+    Hadron = 81,
```
sudojan, line 106: __Codes 81–100 are reserved for generator-specific pseudoparticles and concepts.__
I like, that you keep the PDG code style 👍

Issue #124, opened 2021-02-16
RealPhotonAssumption: Integration and interpolation problems for hard_component


Jean1995:
The calculation of the hard component of the photonuclear crosssection is done by interpolating tabulated values (see [paper by Bugaev et al](https://doi.org/10.1016/j.astropartphys.2004.03.002) or one of the PROPOSAL publications for more details).
However, these tabulated values have a limited definition range (10^-7 < v < 1 and 10^2 GeV < E < 10^9 GeV). Outside of this kinematic range, the hard component is set to zero, causing a discontinuity in the differential cross section.

This causes numerical problems in two cases:

### For 10^5 MeV

For E = 10^5 MeV, this causes a discontinuity in dEdx and dNdx which is poorly interpolated.
![image](https://user-images.githubusercontent.com/15159319/108096785-a9fdb280-7081-11eb-98dc-8e7f7b6d1a43.png)
In the direct vicinity of 10^5 MeV, we observe deviations that are 10% at most.

### For high energies (E > 10^10 MeV)

For higher energies, the kinematic lower limit in v gets smaller and will at some point start to include the discontinuity in v.
The current integration method in PROPOSAL becomes unstable when integrating over the discontinuity in the differential cross section, which causes oscillations in dEdx and dNdx for high energies:

![image](https://user-images.githubusercontent.com/15159319/108097478-753e2b00-7082-11eb-945b-bde5e45d4c2b.png)

Above 10^10 MeV, we expect the difference between the interpolated value and the integrated value to be below 1%. 

**Possible solutions?**

The first issue may be solved by avoiding the hard cutoff for E=10^5 MeV

The second issue may be solved by either using a more robust integration method or by solving the hard cutoff in v.

Jean1995:
Apparently, the integration problems at high energies (when using small energy cuts such an e_cut of 500 MeV) are also present when not using the hard_component, although they are less severe in this case.

This for example is the dNdx interpolation precision for muons in ice, using an ecut of 500 MeV and using the Zeus parametrization:

![image](https://user-images.githubusercontent.com/15159319/151367441-a9c7516e-ad82-4cc5-810c-bb1b39b98e91.png)

We see that we get interpolation problems for energies about 10^10 MeV. If we zoom in on the dNdx values, we see "jumps" in dNdx again:

![image](https://user-images.githubusercontent.com/15159319/151367728-752a15bc-65ff-4dfd-83e0-f6fb0cdefc5f.png)

[...]

PR #202, opened 2021-09-15
Use empirical correction factor for BremsElectronScreening only for energies below 50 MeV

Jean1995:
The parametrization `BremsElectronScreening` is supposed to reproduce the behavior of the bremsstrahlung crosssetion used in [EGS](https://www.researchgate.net/publication/255277661_The_EGS5_code_system).

However, in EGS, the empirical correction factor `A` is only applied for energies above 50 MeV. This is ok for higher energies because the Columb correction is treated numerically using the Colulomb correction term `f_c`.

However, in PROPOSAL, we used the correction factor `A` also for energies above 50 MeV, and for energies above 10 GeV, we used the correction factor at 10 GeV. This is wrong because this basically accounts for the Coulomb correction twice. The difference in the crosssection is going to be a few percents for high energies.

### DIFF
--- a/src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/Bremsstrahlung.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/Bremsstrahlung.cxx
@@ -496,9 +496,6 @@ double crosssection::BremsElectronScreening::CalculateParametrization(
         phi2 = phi1;
     }
 
-    // Coulomb correction function and empirical correction factor
-    double A_fac = interpolant_->InterpolateArray(logZ, energy);
-
     aux = ALPHA * comp.GetNucCharge();
     aux *= aux;
     auto f_c = aux
@@ -541,8 +538,11 @@ double crosssection::BremsElectronScreening::CalculateParametrization(
 
     xi = Lp / (Lr - f_c);
 
+    // Coulomb correction function and empirical correction factor
+    double A_fac = 1.0;
     if (energy < 50.) {
         f_c = 0;
+        A_fac = interpolant_->InterpolateArray(logZ, energy);
     }
 
     auto result = A_fac * comp.GetNucCharge() * (comp.GetNucCharge() + xi);


Jean1995:
Two plots to show the difference due to the correction (dEdx/E for electrons in air, only continuous losses):

test1

test2

You can see the difference quite well in the ratio plot: Until 50 MeV, there is no difference. Then we get a hard jump in the crosssection. Afterwards, there are some changes in the crosssection until 10 GeV, where the correction factor used in the old version starts to be fixed.

sudojan:
So this fix introduces a discontinuity in the cross section?
I assume the CubicSpline Interpolation will get its issues with this kink, as it is the case for the photonuclear interaction, see #124 
Would a smooth transition be more realistic?

maxnoe:
@Jean1995 Can we extract the values from EGS somehow and directly compare?

Jean1995:
> So this fix introduces a discontinuity in the cross section?
> I assume the CubicSpline Interpolation will get its issues with this kink, as it is the case for the photonuclear interaction, see #124

Comparing integration and interpolation, there are some slight oscillations visible, but the error does not get too high:

test3

test4

So the interpolation routine basically introduces a smooth transition by itself.

This discontinuity has been the reason why I did not want to use this hard cutoff when I implemented the crosssection in the first place. I've been thinking about a smooth transition as well, but it felt rather arbitrary to me to introduce some kind of smoothing. One way would be to set all correction factors above 50 MeV to '1', so that the interpolation of the correction factors is naturally smoothed out a bit, but I believe the result will be pretty similar to the results we already see above (including the oscillations).

sudojan (approved):

sudojan:
Indeed, the interpolation is doing some sort of smooting intrinsically.
However, this depends of course on the chosen settings.
As long as this is documented somewhere, this PR can be approved.

sudojan:
> 
> 
> @Jean1995 Can we extract the values from EGS somehow and directly compare?

This comparison is still something to look at, of course.

Jean1995:
> > @Jean1995 Can we extract the values from EGS somehow and directly compare?
> 
> This comparison is still something to look at, of course.

I'm currently trying to get EGSnrc / PEGS4 running, this may also be interesting for other comparisons

asandrock:
Introducing a kind of smoothing would actually be not too dissimilar to the treatment available as second option in EGSnrc: besides the standard treatment from EGS4 (Born cross section with emipirical correction factor below 50 MeV, asymptotic high-energy cross section above that), there is now a second option which uses the asymptotic cross section above 50 MeV, partial wave calculations by Tseng & Pratt below 2 MeV, and a spline interpolation for 2 MeV < E < 50 MeV, and a third one based on tabulated data.

Extracting the energy loss from EGS4 is not straightforward, as far as I can see. Using EGSnrc for simulations seems to be not too difficult, but how to get to the implemented cross sections or the corresponding energy loss seems to be far from obvious.

PR #205, opened 2021-09-21
Fixes and improvements for calculation of ionzation secondaries

Jean1995:
There are two improvements in the calculation of secondary particles from ionization losses:

Firstly, there has been a fix in the energy calculation of the delta electrons where the initial rest mass of the atomic electron has been ignored so far.

Secondly, the directions of the secondary particles are now sampled according to four-momentum-conservation.
To ensure that the sampling procedure is physically consistent, I added a UnitTest checking the momentum and energy conservation for secondaries::NaivIonization. More UnitTests, also for other secondary calculators, can and should be added in the future.


src/PROPOSAL/detail/PROPOSAL/secondaries/parametrization/ionization/NaivIonization.cxx
```diff
@@ -14,24 +14,42 @@ using namespace PROPOSAL;
...
+    dir_delta.deflect(cosphi_bar, std::fmod(rnd * 2. * PI + PI, 2. * PI));
+
+    return make_tuple(dir_outgoing, dir_delta);
 }
 
 std::tuple<double, double> secondaries::NaivIonization::CalculateEnergy(
     double energy, double v)
 {
-    return make_tuple(energy * (1 - v), energy * v);
+    return make_tuple(energy * (1 - v), energy * v + ME);
```
sudojan, line 43: why adding the electron mass?

sudojan: so the calculated/sampled relative energy loss in the Ionization does not stand for the total energy, like for the other cross sections, but the kinetic energy?

PR #207, opened 2021-09-29
Fix bug in calculation of photonuclear secondaries

Jean1995:
Fix wrong assignment of particle properties in secondaries::Photonuclear::CalculateSecondaries

A bug introduced MR #199 caused the wrong properties to be assigned to the lepton, and the hadron has been completely uninitialized. This bug is therefore present in version 7.0.7

### DIFF
--- a/src/PROPOSAL/PROPOSAL/secondaries/parametrization/photonuclear/Photonuclear.h
+++ b/src/PROPOSAL/PROPOSAL/secondaries/parametrization/photonuclear/Photonuclear.h
@@ -31,12 +31,12 @@ namespace secondaries {
             primary_lepton.propagated_distance = 0.;
 
             auto hadron = ParticleState();
-            primary_lepton.energy = loss.energy;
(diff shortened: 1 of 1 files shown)


src/PROPOSAL/PROPOSAL/secondaries/parametrization/photonuclear/Photonuclear.h
```diff
@@ -31,12 +31,12 @@ namespace secondaries {
             primary_lepton.propagated_distance = 0.;
 
             auto hadron = ParticleState();
```
maxnoe, line 33: This does an unnecessary copy, correct?
```
ParticleState hadron;
```

maxnoe: Yes, you're right. Still, I think `ParticleState hadron;` is the canonical form, shorter and thus more readable.

PR #210, opened 2021-10-11
Fix numerical issues when propagating high energy muons with low energy cut settings

Jean1995:
There is a bug in the current version of PROPOSAL when trying to propagate high energy particles with a relatively low energy cut (thanks to @wjwoodley for bringing this to our attention!)

### Bug description

In this case, muons have been propagated in StandardRock for a distance of 1 km, using an e_cut of 500 MeV. Looking at the final particle energies for different initial energies, the histogram looked like this:
old

With this fix, the histogram changes to this 
new
which is in agreement with PROPOSAL 6 results.

### Bug source

The source of this bug has been the sampling of the energy of the next stochastic interaction ('InteractionBuilder::EnergyInteraction()'). For random numbers close to one (i.e. the edge of the parameter space), the energy of the next stochastic interaction should be very close to the initial energy. Instead, the interaction energy is quite far away from the initial energy, causing an nonphysical, very high continuous loss. 
Normally, this bug should only occur in very rare cases. However, when using high initial energies with a low e_cut, there are a lot of steps during the propagation process, so that this nonphysical case starts to appear in almost every propagation.

### Bug solution

The bug occurs because the NewtonRaphson method of CubicInterpolation is unreliable in this very specific case. Changing the accuracy or the number of steps does not improve the results. However, the results have been vastly improved by changing the initial condition of the NewtonRaphson method: Normally, the initial conditions have been determined by the CubicInterpolation library by performing a bisection search to find an interval that includes only 1% of the original parameter space, and use the middle of this interval as a starting point for the NewtonRaphson method. But if we are very close to the original upper_limit of the problem, it seems to be much better to use the upper_limit as an initial guess of the NewtonRaphon method. 

A similar procedure has been used in former versions of PROPOSAL: If the random number has been very close to the edge of the parameter space, only two NewtonRaphson steps, starting at the edge of the parameter space, have been performed (see [here](https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/37059e4b6337b3db908e33580eb2795c5473a196/private/PROPOSAL/propagation_utility/PropagationUtilityInterpolant.cxx#L318))

To test this specific case, I also adapted the UnitTests to account for random numbers close to the edge of the parameter space.

### DIFF
--- a/src/PROPOSAL/detail/PROPOSAL/propagation_utility/PropagationUtilityInterpolant.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/propagation_utility/PropagationUtilityInterpolant.cxx
@@ -93,8 +93,21 @@ double UtilityInterpolant::GetUpperLimit(double upper_limit, double rnd)
 
     auto integrated_to_upper = interpolant_->evaluate(upper_limit);
     auto initial_guess = cubic_splines::ParameterGuess<double>();
-    initial_guess.x = NAN;
-    initial_guess.upper = upper_limit;
+
+    // find initial parameters for newton raphson method by using bisection
+    auto f = [this, &integrated_to_upper, &rnd](double val) {
+        return interpolant_->evaluate(val) - (integrated_to_upper - rnd);
+    };
+    auto bisec_tolerance = (upper_limit - lower_lim) * 1e-2;
+    std::tie(initial_guess.lower, initial_guess.upper) =
+            Bisection(f, lower_lim, upper_limit, bisec_tolerance, 100);
+
+    // if we are close to the upper limit, start newton raphson method there
+    if (initial_guess.upper == upper_limit) {
+        initial_guess.x = upper_limit;
+    }
+    else
+        initial_guess.x = (initial_guess.lower + initial_guess.upper) / 2;
 
     try {
         return cubic_splines::find_parameter(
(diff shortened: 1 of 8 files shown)


src/PROPOSAL/detail/PROPOSAL/propagation_utility/PropagationUtilityInterpolant.cxx
```diff
@@ -93,8 +93,21 @@ double UtilityInterpolant::GetUpperLimit(double upper_limit, double rnd)
...
+    };
+    auto bisec_tolerance = (upper_limit - lower_lim) * 1e-2;
+    std::tie(initial_guess.lower, initial_guess.upper) =
+            Bisection(f, lower_lim, upper_limit, bisec_tolerance, 100);
+
+    // if we are close to the upper limit, start newton raphson method there
+    if (initial_guess.upper == upper_limit) {
+        initial_guess.x = upper_limit;
+    }
+    else
```
sudojan, line 109: Could the lower limit also be problematic?
Since we don't know that, at least a comment, here, would be helpful.

Jean1995:
> Is this only an issue for the PropagationUtility or also for the cross section integrals? For the cross sections, always the lower limit is used. However, for the propagation integrals, the mid or the upper limit is used. This seems inconsistent. A comment (also in the code) would be helpful to understand.

For the cross section interpolation, the Bisection method of PROPOSAL is actually used to find an energy where the cross section is already zero, but not the exact energy. I use the lower limit here because this basically reproduced the same behavior that we had before.

PR #215, opened 2021-10-19
Change default photopairproduction crosssection for Gammas to KochMotz

Jean1995:
Basically, this was just forgotten in the last release.

This is more consistent with the default crosssections used for electrons and positrons (ElectronScreening).


sudojan:
This should be documented somewhere.

PR #222, opened 2021-11-08
Add Photoproduction, CrossSectionDirect and ParametrizationDirect

Jean1995:
This PR adds interactions of photons with atomic nuclei, 'Photoproduction', to PROPOSAL.

Since these parametrizations are not differential, but given in an analytical form, a CrossSection class
('CrossSectionDirect') and a corresponding Parametrization class ('ParametrizationDirect') have been added.

These classes can be used when the parametrization objects is completely responsible for calculating the cross sections. This can be useful if cross sections are, like in this case, given in an analytical form, but other use cases are also possible (e.g. where cross sections are given as tabulated values).


src/PROPOSAL/PROPOSAL/crosssection/CrossSectionDirect.h
```diff
@@ -0,0 +1,61 @@
...
+
+#include "PROPOSAL/crosssection/CrossSection.h"
+#include "PROPOSAL/crosssection/parametrization/ParametrizationDirect.h"
+
+namespace PROPOSAL {
+    class CrossSectionDirect : public CrossSectionBase {
+    public:
+        template <typename Param,
+                typename _name = crosssection::ParametrizationName<Param>>
+        CrossSectionDirect(Param param_, ParticleDef p,
```
sudojan, line 11: The design concept, here, is something, I think, can be improved. But I would like a broader discussion about that.

PROPOSAL initially only had interactions differential in the energy loss, so they always had to be integrated or interpolated for a dEdx or dNdx.
This changed, when the weak interaction or annihilation were introduced, which are only stochastic and don't have continuous losses. However, both interactions still have to be integrated or ininterpolated, since they are given differential  on how the energy is separated between the secondary particles.
To make them adaptable, they got an dEdx=0 or dE2dx=0 and still some energy cuts, which were not used.

Now with this new sort of interaction, the photoproduction, also the interpolation gets obsolet, raising the question if it's time for some restructuring.
There should be a class of only stochastic cross sections with minimal overhead, necessary for this photoproduction. Derived from that, a cross section class still consisting only of stochastic interaction, but which are differential in partitioning the energy to the secondaries and therefor neds to be integrated for the total cross section. (Btw. also the CSMS parametrization of the weak interaction have the total cross sections also tabulated, so they don't need to be integrated, which might be more stable.)

In this approach the interactions, which are able to create continuous energy losses, and the interactions, where the primary particle dosen't exist anymore after the interaction (where also the decay might be considered) gets clearly separated.

ps: this might also help in the long term TODO to perform only-stochastic propagations, e.g. for neutrinos.

src/PROPOSAL/PROPOSAL/crosssection/CrossSectionBuilder.h
```diff
@@ -4,11 +4,14 @@
...
+#include "PROPOSAL/crosssection/CrossSectionDirect.h"
 
 namespace PROPOSAL {
+namespace detail {
 template <typename Param, typename P, typename M>
-auto make_crosssection(Param&& param, P&& p_def, M&& medium,
-    std::shared_ptr<const EnergyCutSettings> cuts, bool interpolate)
+auto make_crosssection_impl(Param&& param, P&& p_def, M&& medium,
+                            std::shared_ptr<const EnergyCutSettings> cuts,
+                            bool interpolate, std::false_type)
```
sudojan, line 14: a hint what this true_type or false_type stands for, would be helpful

src/PROPOSAL/PROPOSAL/crosssection/parametrization/PhotoRealPhotonAssumption.h
```diff
@@ -32,16 +32,14 @@
 #include <memory>
 #include <unordered_map>
 
-#define PHOTO_PARAM_REAL_DEC(param, parent)                                    \
-    class Photo##param : public Photo##parent {                                \
+#define PHOTO_PARAM_REAL_DEC(param)                                            \
+    class Photo##param : public PhotoRealPhotonAssumption {                    \
```
sudojan, line 36: Since the photonuclear cross section (i.e. a charged particle interacting inelastically with a nucleus via a virtual photon) is now better decoupled from the photohadronic interaction (i.e. a real photon interacting with a nucleus, sometimes also called photoproduction), the parametrization to calculate the photonuclear interaction out of the photohadronic interaction in the GVD model (general vector dominance model) can also be modularized. Currently there is just one parametrization (BezrukovBugaev) but there are further parametrizations, like the BorogPetrukhin cross section (used in many applications e.g. in GEANT4), which can then be implemented.

sudojan: Then, there is no PhotoZeus or Rhode parametrization, but a PhotoBezrukovBugaev parametrization getting a PhotoproductionZeus, oder PhotoproductionBezrukovBugaev.
Like the ALLM Parametrizations get a shadowing parametrization.

sudojan (changes requested):
So there are two major changes that I request:
- first is the creation of a vector-meson-dominance photonuclear parametrization independent of the given Photoproduction parametrization. Currently there is just one parametrization (BezrukovBugaev) the user can chose, but e.g. the BorogPetrukhin parametrization used in GEANT4 can then be added fairly easily.
- restructuring of the cross section base classes, so not every stochastic cross section needs to implement a dedx calculation or an interpolation. This of course affects also the propagation utilities and the propgation routines, but would also enable purely stochastic propagations.

However, especially the second requst would introduce major changes and a major release. They would also be too large for a single PR.
Therefore I am satisfied if issues concerning these requests would be created, labed with the PROPOSAL8 milestone.

Jean1995:
> So there are two major changes that I request:
> 
>     * first is the creation of a vector-meson-dominance photonuclear parametrization independent of the given Photoproduction parametrization. Currently there is just one parametrization (BezrukovBugaev) the user can chose, but e.g. the BorogPetrukhin parametrization used in GEANT4 can then be added fairly easily.
> 
>     * restructuring of the cross section base classes, so not every stochastic cross section needs to implement a dedx calculation or an interpolation. This of course affects also the propagation utilities and the propgation routines, but would also enable purely stochastic propagations.
> 
> 
> However, especially the second requst would introduce major changes and a major release. They would also be too large for a single PR. Therefore I am satisfied if issues concerning these requests would be created, labed with the PROPOSAL8 milestone.

I agree with both requests. However, the first request would also break backwards compatibility and therefore require a major release (since people could, for example, not use a `Zeus` parametrization anymore, but they would need to create a `BezrukovBugaev` parametrization which gets a `Zeus` parametrization as a parameter. I am completely convinced that this is a useful change, but I would also wait until the next major release for that.

Jean1995:
I've been putting the change requests to a [discussion](https://github.com/tudo-astroparticlephysics/PROPOSAL/discussions/231#discussion-3720904) instead of creating a new issue.
I am not sure whether this is the "better" way of collecting the ideas, but I kind of like the idea to collect them in one place.

Jean1995:
Do we want to include the photoproduction cross section that has been used in CORSIKA 7?

It is described on page 48 of [FZKA 6019](https://web.iap.kit.edu/corsika/physics_description/corsika_phys.pdf) as a continuum with three resonances. However, I am not sure how these resonances can be implemented as a cross section. I believe that for EGS4, all data has only been provided as a fit to data.

sudojan:
> Do we want to include the photoproduction cross section that has been used in CORSIKA 7?
> 
> It is described on page 48 of [FZKA 6019](https://web.iap.kit.edu/corsika/physics_description/corsika_phys.pdf) as a continuum with three resonances. However, I am not sure how these resonances can be implemented as a cross section. I believe that for EGS4, all data has only been provided as a fit to data.

This would be similar to the Rhode parametrization, where also some empirical data get interpolated.
But this is not necessary for this PR, but can also be included if it's not a big deal.

sudojan (approved):

Issue #221, opened 2021-11-01
Information about interaction target in SecondariesVector missing


Jean1995:
If we want to calculate a list of `ParticleState` from a `StochasticLoss` object after propagation, for example in the case of [MuonPairProduction](https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/3bda79a4e05280b09def6066f9594d27923ede29/src/PROPOSAL/PROPOSAL/secondaries/parametrization/mupairproduction/KelnerKokoulinPetrukhinMupairProduction.h#L33), we need the information about the target (Component) that we interacted with. However, this information is missing in the output of our Propagator (`Secondaries`), so it is not possible to use this function.

We need to add this information for each StochasticInteraction.

Jean1995:
In what form should we include the information about the target that we interacted with?

My first idea would be to include the hash of the target. If we interacted with a Component, we include the hash of the Component (where we can trace back the Component object using the component_map). If we interacted with a Medium (i.e. if we have Ionization), we could include the hash of the medium. However, there is no way at the moment to trace back the original medium yet.

Jean1995:
Solved by #224

PR #224, opened 2021-11-12
Add target_hash to StochasticLosses

Jean1995:
This adds information about the component that we interacted with to the Secondaries object and the StochasticLoss object.
If we interacted with a component, we can use the `GetComponent` method to directly access the Component. This can then be used to sample secondary particles (solves issue #221).

In a future release, we might want to replace the hash with a shared_ptr to the original Component since this is more transparent for the user. However, we would need to restructure some parts of  the `Interaction` (and `Interaction::Loss`) and also `CrossSection` classes since all of these classes work with the hashes instead of pointer.


src/PROPOSAL/detail/PROPOSAL/particle/Particle.cxx
```diff
@@ -125,10 +126,20 @@ double ParticleState::GetMomentum() const
...
                                double propagated_distance,
-                               double parent_particle_energy)
+                               double parent_particle_energy, size_t target_hash)
                                : Loss(type, loss_energy, parent_particle_energy),
                                position(position), direction(direction), time(time),
-                               propagated_distance(propagated_distance) {}
+                               propagated_distance(propagated_distance),
+                               target_hash(target_hash) {}
+
+Component StochasticLoss::GetComponent() const {
```
sudojan, line 135: How to get the medium?

sudojan (changes requested):
An example script or Unit Test is missing.

src/PROPOSAL/detail/PROPOSAL/particle/Particle.cxx
Jean1995: I added a `medium_map` to the `Medium` class to get the corresponding Medium to a hash.
The `GetComponentForHash()` method has also been moved to be a static function of the Component class, there is also a `GetMediumForHash()` method now in the Medium class.

I think for the future, we should use pointers instead of these global maps.

Jean1995:
> An example script or Unit Test is missing.

I added an example on how to extract the StochasticLosses and the information about the target hash/medium to the Propagator jupyter notebook example.

I also added UnitTest for the MediumMap and the ComponentMap to check whether all objects have been correctly registered.

src/PROPOSAL/detail/PROPOSAL/particle/Particle.cxx
sudojan: > I think for the future, we should use pointers instead of these global maps.

well, then create an issue

sudojan:
> The ID of the Interaction Type is 1000000004. The hash of the target that we interacted with is 9764564051507188568.

In the example notebook of the propagator, this output could be a little bit more descriptive. How to get component/medium name out of the hash is already shown afterwards, but this is not done for the interaction type. Although this might already be shown in another example notebook, it would be practical to have this also here.

src/PROPOSAL/PROPOSAL/Secondaries.h
```diff
@@ -51,11 +51,13 @@ class Secondaries {
 
     // Operational functions to fill and access track
     void reserve(size_t number_secondaries);
-    void clear() { track_.clear(); types_.clear(); };
-    void push_back(const ParticleState& point, const InteractionType& type);
+    void clear() { track_.clear(); types_.clear(); target_hashes_.clear(); };
+    void push_back(const ParticleState& point, const InteractionType& type,
+                   const size_t& target_hash = 0);
```
sudojan, line 56: Why does the hash get a default value?

Jean1995: Mainly for backward compatibility for people who might have used the `push_back` function without specifying a target_hash

Jean1995:
> > The ID of the Interaction Type is 1000000004. The hash of the target that we interacted with is 9764564051507188568.
> 
> In the example notebook of the propagator, this output could be a little bit more descriptive. How to get component/medium name out of the hash is already shown afterwards, but this is not done for the interaction type. Although this might already be shown in another example notebook, it would be practical to have this also here.

Done. This was a very good idea, since it was not even clear to me how to extract this information in python 🙈

examples/Propagator.ipynb
```diff
@@ -281,6 +281,123 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We can see that our `StochasticLoss` objects carries the information about the target that we interacted with in the form of a hash.\n",
+    "For Ionization interactions, this corresponds to the hash of a Medium object.\n",
+    "For all other interactions, this corresponds to the hash of a Component object (i.e. an atom).\n",
+    "We can get the Medium/Component object corresponding to these hashes using the methods `pp.medium.medium_map()` and `pp.component.component_map()`:"
```
sudojan, line 366: Due to the recent changes, this comment should refer to the methods `pp.medium.get_medium_for_hash()` and `pp.component.get_component_for_hash()`

sudojan (approved):

PR #226, opened 2021-11-21
Disable caches in GitHub actions and fix Secondaries_TEST

Jean1995:
Caching the build in GitHub actions seems to cause some problems. Sometimes, the wrong cache is loaded, so the test are sometimes run with an outdated build.
I disabled the caches and just combined building and testing into one run for now.

Furthermore, there have been some floating point precision in Secondaries_TEST on Windows. This PR fixes them.

### DIFF
--- a/.github/workflows/cpp.yml
+++ b/.github/workflows/cpp.yml
@@ -24,6 +24,7 @@ jobs:
       env:
         CC: ${{ matrix.compiler }}
         CXX: ${{ matrix.compilerpp }}
+        PROPOSAL_TEST_FILES: ${{ github.workspace }}/PROPOSAL_BUILD/tests/TestFiles
       steps:
         - uses: actions/checkout@v2
         - name: Install gcc-5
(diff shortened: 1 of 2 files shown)


sudojan:
Is this a temporary PR until the Issue loading an outdated cache is fixed? Is there an Issue filed?

maxnoe:
If the issue is outdated conan definitions, add the hash of the conanfile to the cache key

LarsBollmann:
I'm not an expert on this, so take this with a grain of salt.

I think the problem could be the same cache key for different builds.

According to the [documentation](https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows#using-the-cache-action):
> Once you create a cache, you cannot change the contents of an existing cache but you can create a new cache with a new key.

So if a cache already exists and you build a new version, then the new build is not cached with that key.
This causes the tests to be potentially executed with outdated build caches.

For the conan cache a hashed conanfile as part of the key would solve that problem as Max already suggested. 
`${{ runner.os }}-cache-conan-${{ matrix.compiler }}-${{ hashFiles('conanfile.py') }}-key`

In the build cache you could add the commit hash.
`${{ runner.os }}-cache-build-${{ matrix.compiler }}-${{ github.sha }}-key`

If the build cache is needed between different workflow runs (e.g. to increase build speed, which I'm not entirely sure is the case here) you could use the following to the cache command
```
 - name: Cache build
          id: cache-build
          uses: actions/cache@v2
          with:
            path: PROPOSAL_BUILD
            key: ${{ runner.os }}-cache-build-${{ matrix.compiler }}-${{ github.sha }}-key
            restore-keys: |
              ${{ runner.os }}-cache-build-${{ matrix.compiler }}-
 ```
 which loads the latest cache with a key starting with the given prefix under 'restore-keys' but still saves a new cache with the given key in the 'key' option.

Jean1995:
I believe the fact that the commit id has not been included in the key was the reason why the wrong build has been used.

I included the improvement of @LarsBollmann in PR #228 which should, if approved, replace this PR.

sudojan:
Since this has been fixed in the already merged PR #228 this can be closed

PR #228, opened 2021-11-23
Fix caches in GitHub actions and fix Secondaries_TEST

Jean1995:
Follow-up to #226

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -12,7 +12,7 @@ jobs:
           uses: actions/cache@v2
           with:
             path: ~/.conan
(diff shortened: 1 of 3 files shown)


LarsBollmann (approved):

.github/workflows/python.yml
```diff
@@ -12,7 +12,7 @@ jobs:
           uses: actions/cache@v2
           with:
             path: ~/.conan
-            key: ${{ runner.os }}-cache-conan-gcc
+            key: ${{ runner.os }}-cache-conan-${{ matrix.compiler }}-${{ hashFiles('conanfile.py') }}-key
```
Jean1995, line 15: matrix.compiler should be gcc here, right?

PR #234, opened 2021-12-21
Change setup.py to be able to find correct python versions

Jean1995:
When trying to install proposal with pip in virtual environments, I realized that python does not find the installed modules.

Looking at the verbose output of `pip install`, CMake said:
```
-- Found Python: /usr/bin/python3.10 (found version "3.10.1") found components: Interpreter Development Development.Module Development.Embed
```
which means that CMake does not use the python version of the virtual environment, but rather the global python version!
The CMake output also said
```
CMake Warning:
    Manually-specified variables were not used by the project:

      PYTHON_EXECUTABLE
      PYTHON_INCLUDE_DIR
      PYTHON_LIBRARY
```
so the variables from the `setup.py` that should set the correct python version are in fact ignored!

From the [manual page of `FindPython`](https://cmake.org/cmake/help/git-stage/module/FindPython.html), I learned that we can set `Python_EXECUTABLE` as an artifacts specification. If I do that (as it is done by the changes in this PR draft), pypi finds the correct python version and everything seems to work smoothly.

(I tested this on both MacOS (CMake 3.20.3) and Arch Linux (3.22.1))

However, I am not sure if this is the correct way of solving this problem. If I understand the manual page correctly, `Python_EXECUTABLE` has only been introduced in CMake 3.16, so I am not sure what will happen for older CMake 
versions. 

@maxnoe, maybe you could have a short look at this issue?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,10 +99,6 @@ class build_ext_cmake(build_ext):
         cmake = get_cmake()
 
         rpath = '@loader_path' if sys.platform == 'darwin' else '$ORIGIN'
-        python_lib = pathlib.Path(
-            sysconfig.get_config_var('LIBDIR'),
-            sysconfig.get_config_var('INSTSONAME')
-        )
         CMAKE_CXX_FLAGS = ""
         if not os.getenv("NO_CONAN", False):
             print(
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ def get_cmake():
 
         if ret.returncode == 0:
             return exe
-    raise OSError("You need cmake >= 3.9")
+    raise OSError("You need cmake >= 3.16")
 
 
 def exists_conan_default_file():
--- a/setup.py
+++ b/setup.py
@@ -132,12 +132,10 @@ class build_ext_cmake(build_ext):
             '-DBUILD_PYTHON=ON',
             '-DCMAKE_POSITION_INDEPENDENT_CODE=TRUE',
             '-DBUILD_EXAMPLE=OFF',
-            '-DPYTHON_EXECUTABLE=' + sys.executable,
-            '-DPYTHON_LIBRARY=' + str(python_lib),
+            '-DPython_EXECUTABLE=' + sys.executable,
             '-DCMAKE_INSTALL_RPATH={}'.format(rpath),
             '-DCMAKE_BUILD_WITH_INSTALL_RPATH:BOOL=ON',
             '-DCMAKE_INSTALL_RPATH_USE_LINK_PATH:BOOL=OFF',
-            '-DPYTHON_INCLUDE_DIR=' + sysconfig.get_path('include')
         ]
         if CMAKE_CXX_FLAGS:
             cmake_call.append(CMAKE_CXX_FLAGS)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires = [
     "setuptools>=45",
     "wheel",
-    "cmake>=3.9",
+    "cmake>=3.16",
     "conan>=1.33",
 ]
 build-backend = "setuptools.build_meta"


maxnoe:
As far as I understand it, `Python_Executable` is related to calling `find_package(Python ....)` as opposed to relying on pybind11 to find the python.

So if we want to use the modern `FindPython`, we should add that find_package call. See

https://pybind11.readthedocs.io/en/stable/compiling.html#findpython-mode

Jean1995:
> So if we want to use the modern `FindPython`, we should add that find_package call. See
> 
> https://pybind11.readthedocs.io/en/stable/compiling.html#findpython-mode

But aren't we already doing that in the [CMakeLists.txt](https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/f64c202ecec486c73077d6e9d9ece0f3e19e87a0/src/pyPROPOSAL/CMakeLists.txt#L5)?

maxnoe:
@Jean1995 Right, then it is clear that the lowercase versions should have been used already...

maxnoe:
I think we should keep setting all, looking at:
https://cmake.org/cmake/help/latest/module/FindPython.html#artifacts-specification

and require cmake 3.16?

Jean1995:
Only specifying `Python_EXECUTABLE` worked on both my Linux and my MacOS machine.

If I specify `Python_EXECUTABLE`, `Python_LIBRARY` and `Python_INCLUDE_DIR` the way it has been done in the previous setup.py, it worked on Arch, but it did not work on MacOS since `Python_LIBRARY` was resolved to `/usr/local/opt/python@3.9/Frameworks/Python.framework/Versions/3.9/lib/Python.framework/Versions/3.9/Python` (which is a non-existing path) instead of (what I believe should be) `/usr/local/Cellar/python@3.9/3.9.7/bin/python3.9`

Jean1995:
Not sure whether there is a platform-independent way to reliably set all these paths. Maybe only setting the Python_EXECUTABLE is enough? Or could this lead to inconsistencies?

Jean1995:
Since this seems to work, I would propose to just set the Python_EXECUTABLE.

Because CMake 3.16 should only be required if one uses the setup.py, it should be enough to alter the requirement in the pyproject.toml file.

maxnoe:
> If I specify Python_EXECUTABLE, Python_LIBRARY and Python_INCLUDE_DIR the way it has been done in the previous setup.py, it worked on Arch, but it did not work on MacOS since Python_LIBRARY was resolved to /usr/local/opt/python@3.9/Frameworks/Python.framework/Versions/3.9/lib/Python.framework/Versions/3.9/Python (which is a non-existing path) instead of (what I believe should be) /usr/local/Cellar/python@3.9/3.9.7/bin/python3.9

`Python_LIBRARY` has to be the actual library, i.e.  `/path/to/libpython3.9.so` on linux and `/path/to/libpython.dylib???` on macos.

maxnoe:
is seems `instsoname` is not what I thought it would be. `sysconfig.get_config_var('LIBDIR')` is the path with `libpython3.9.dylib` inside on my macvm, but `INSTSONAME` is Python.framework/Versions/3.9/Python and not `libpython3.9.dylib`

maxnoe:
I went through the sysconfig variables and sys attributes again and didn't find anything that works on both linux and macos.

If it works with only giving executable, let's do that and hope it does not create another issue.

setup.py
```diff
@@ -132,12 +132,10 @@ def build_cmake(self, ext):
             '-DBUILD_PYTHON=ON',
             '-DCMAKE_POSITION_INDEPENDENT_CODE=TRUE',
             '-DBUILD_EXAMPLE=OFF',
-            '-DPYTHON_EXECUTABLE=' + sys.executable,
-            '-DPYTHON_LIBRARY=' + str(python_lib),
```
maxnoe, line 136: Please remove the now unused code above (`pythonlib = ...`)

Jean1995: Done

maxnoe (approved):

PR #240, opened 2022-01-07
Improve v node distribution for ionization interpolant and fix test issues

Jean1995:
While doing an overhaul of the UnitTest, I discovered that the Ionization stochastic losses have not been tested due to a misplaced `break` instruction. When enabling these tests, they failed for specific random numbers, so I had a closer look.

The following plots are for muons in ice, with an EnergyCut of 500 MeV and an initial energy of 1e11 MeV. The parametrization is BetheBlochRossi.

Looking at the ionization losses v sampled for a specific random number (rnd between 0 and 1), they look like this:

plot1_master-1

If we look at the ratio of the sampled v values for the integral and interpolant, we see this pattern:

plot2_master-1

which means that the losses are off by almost 2 percent.

If we look at the cumulative cross section (i.e. the integral of the differential cross section from v_cut to a specific v), together with the nodes of our interpolation table, we get this image:

plot3_master-1

From this, we can see that there are not enough nodes for small v to describe the cumulative cross section with sufficient accuracy.

Therefore, I tweaked the function which described the distribution of our nodes in v so that there are more nodes for small v. The cumulative cross section with this PR looks like this:

plot3_pr-1

This also improves the ratio plot above, which now looks like this:

plot2_pr-1

This example was for 1e11 MeV, which is a rather high energy where ionization losses are not dominant. While the discrepancies are larger for high energies, they are also visible for smaller energies.
For example, for 1e5 MeV, the ratio plots showing the stochastic losses looks like this:

plot2_master

The error is now "only" 0.5% for a smaller range of random numbers, but still visible. 

In my opinion, these discrepancies justify this PR. As an alternative, we could accept that there are uncertainties for certain random numbers, especially at high energies.

### DIFF
--- a/src/PROPOSAL/PROPOSAL/crosssection/CrossSectionDNDX/CrossSectionDNDXInterpolant.h
+++ b/src/PROPOSAL/PROPOSAL/crosssection/CrossSectionDNDX/CrossSectionDNDXInterpolant.h
@@ -11,44 +11,52 @@
 
 namespace PROPOSAL {
 
-double transform_relative_loss(double v_cut, double v_max, double v);
-double retransform_relative_loss(double v_cut, double v_max, double v);
+double transform_relative_loss(double v_cut, double v_max, double v, double c = 1);
+double retransform_relative_loss(double v_cut, double v_max, double v, double c = 1);
 double transform_loss_log(double v_cut, double v_max, double v);
 double retransform_loss_log(double v_cut, double v_max, double v);
 
-template <typename Param> struct transform_loss {
-    static std::function<double(double, double, double)> func;
-};
-
-template <typename Param> struct retransform_loss {
-    static std::function<double(double, double, double)> func;
-};
-
-template <typename Param>
-std::function<double(double, double, double)> transform_loss<Param>::func
-    = [](double v_cut, double v_max, double v) {
-          return transform_relative_loss(v_cut, v_max, v);
-      };
-
-template <typename Param>
-std::function<double(double, double, double)> retransform_loss<Param>::func
-    = [](double v_cut, double v_max, double v) {
-          return retransform_relative_loss(v_cut, v_max, v);
-      };
-
 namespace crosssection {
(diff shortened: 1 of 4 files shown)


src/PROPOSAL/PROPOSAL/crosssection/CrossSectionDNDX/CrossSectionDNDXInterpolant.h
```diff
@@ -11,44 +11,52 @@
...
+    return retransform_relative_loss(v_cut, v_max, v);
+}
 
-namespace crosssection {
-    struct ComptonKleinNishina;
+// specifications for Ionization
+
+template <typename Param, std::enable_if_t<std::is_base_of<crosssection::Ionization, Param>::value, bool> = true>
+double transform_loss(double v_cut, double v_max, double v) {
+    return transform_relative_loss(v_cut, v_max, v, 1.5);
```
sudojan, line 41: Can you explain, why this additional exponentiation of 1.5 during transformation decreases the interpolation error?

sudojan:
Can you quantify how much this changes the runtime performance?

src/PROPOSAL/PROPOSAL/crosssection/CrossSectionDNDX/CrossSectionDNDXInterpolant.h
Jean1995: This is what the distribution of interpolation nodes look like with c=1 (so the default distribution):

plot3_master

Note that I plotted only every fifth node here so that it is easier to see the effect. 

And this is what the distribution of the interpolation nodes look like with c=1.5:

plot3_master

There are more nodes for small v now, which is the kinematic range where the interpolated value is actually changing.

Jean1995:
> Can you quantify how much this changes the runtime performance?

There is not change in the runtime, since the number of nodes in v (100) stays the same, only its distribution is changed.

sudojan (approved):

PR #241, opened 2022-01-07
Add function "HitGeometry" to Secondaries

### LINKED ISSUE #238: Function of Secondaries 'DidHitDetector'
Add a function to `Secondaries`, which checks whether a specific Geometry has been hit. Returns a bool.

Jean1995:
Fixes #238 

This adds a function which checks whether a particle track has a hit a certain geometry or not. 
We decided that particles where the propagation ended on the border of a geometry (i.e. the particle did not enter the geometry, but "touched" it) will be counted as a hit.

### DIFF
--- a/tests/Secondaries_TEST.cxx
+++ b/tests/Secondaries_TEST.cxx
@@ -154,6 +154,47 @@ TEST(SecondaryVector, EntryPointExitPointRePropagation)
     EXPECT_TRUE(exit_point->propagated_distance == sphere.GetPosition().GetZ() + sphere.GetRadius());
 }
 
+
+TEST(SecondaryVector, HitDetector) {
+    // define our dummy particle track
+    Secondaries dummy_track(nullptr, std::vector<Sector>{});
+    std::vector<Cartesian3D> positions{ {0, 0, 0}, {0, 0, 10}, {0, 0, 20} };
+
+    for (auto p : positions) {
+        ParticleState p_state;
+        p_state.position = p;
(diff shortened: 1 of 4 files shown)


tests/Secondaries_TEST.cxx
```diff
@@ -154,6 +154,47 @@ TEST(SecondaryVector, EntryPointExitPointRePropagation)
...
+    std::vector<Cartesian3D> positions{ {0, 0, 0}, {0, 0, 10}, {0, 0, 20} };
+
+    for (auto p : positions) {
+        ParticleState p_state;
+        p_state.position = p;
+        p_state.direction = Cartesian3D(0, 0, 1);
+        dummy_track.push_back(p_state, InteractionType::Undefined);
+    }
+
+    // test geometries along propagation axis
```
sudojan, line 170: Some test spheres seem redundant, while I would also suggest testing the other geometries.
Would be interesting to see, if the edge case of propagating at the surface of a box is also handled as expected.

sudojan (changes requested):
pybindings are missing

tests/Secondaries_TEST.cxx
Jean1995: I added some tests for other geometries, although I think that this is rather something that should be tested in the UnitTests of the corresponding geometries.
However, I still think that the border case that you described could potentially be pretty important.

Jean1995:
> pybindings are missing

Done

sudojan (approved):

PR #245, opened 2022-01-14
add NoScattering case to make_multiple_scattering function

Jean1995:
Before this, one could not explicitly choose `NoScattering`as a scattering method in the configuration file.

### DIFF
--- a/src/PROPOSAL/PROPOSAL/scattering/Scattering.h
+++ b/src/PROPOSAL/PROPOSAL/scattering/Scattering.h
@@ -118,7 +118,11 @@ public:
     /**
      * @brief random numbers required for multiple scattering.
      */
-    static constexpr size_t MultipleScatteringRandomNumbers() noexcept { return 4; }
+    size_t MultipleScatteringRandomNumbers() const noexcept {
+        if (m_scatter_ptr)
+            return 4;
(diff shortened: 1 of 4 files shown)


sudojan (commented):
can this also be tested?
there is only
https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/master/tests/Scattering_TEST.cxx#L390
for NoScattering

Jean1995:
> can this also be tested? there is only https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/master/tests/Scattering_TEST.cxx#L390 for NoScattering

We can check that we actually get a nullptr for this option.
We could only check what we actually get no deflections if this would be combined with the PropagationUtility object, but this is currently not the case in this UnitTest.

sudojan:
> > can this also be tested? there is only https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/master/tests/Scattering_TEST.cxx#L390 for NoScattering
> 
> We can check that we actually get a nullptr for this option. We could only check what we actually get no deflections if this would be combined with the PropagationUtility object, but this is currently not the case in this UnitTest.

yeah, the last one is what I meant

sudojan (approved):

PR #247, opened 2022-01-15
Restructure PhotoPairProduction secondary calculation classes

Jean1995:
Using PROPOSAL in CORSIKA 8, we realized that the default cross section for PhotopairProduction (`KochMotz`) is incompatible with the default secondary calculator for PhotopairProduction (`Tsai`). 
Therefore, I added a `KochMotz` secondary calculator. To do this in an efficient way, I restructured the `secondaries::PhotoPairProduction` classes a bit.

There are now three parametrizations which can be used to sample the PhotopairProduction secondaries:

1. `PhotoPairProductionTsaiForwardPeak`: Calculate `rho` (energy asymmetry parameter) using the Tsai parametrization. Use the EGS4 approximation to sample the deflection angle.
2. `PhotoPairProductionTsai`: Calculate `rho` using the Tsai parametrization. Use the Tsai parametrization to sample the deflection angle.
3. `PhotoPairProductionKochMotzForwardPeaked` (new default): Calculate `rho` using the KochMotz parametrization. Use the EGS4 approximation to sample the deflection angle.

More parametrization based on KochMotz, using more sophisticated ways to sample the deflection angles, may be added in the future if needed.

Since the calculation of `rho` is similar for all parametrizations, I added a virtual class `PhotoPairProductionInterpolant`. This way, redundant code is avoided.


Jean1995:
Todo list for this PR:

- [ ] Write some basic UnitTest for all parametrizations
- [ ] Adapt documentation

sudojan:
Maybe, there is something I don't get, but why are the description of the energy part and the deflection part combined into one parametrization class. E.g. for pair production, the energy loss or asymmetry is calculated independent of the deflection (of course the latter depends on the energies sampled before) in two separated parametrizations.
So why don't we split the PhotoPairProduction Secondary parametrization into one for the energy asymmetry and one for the angles?
Thereby, the number of parametrizations can be kept to a minimum with no need to implement all combinations.

Jean1995:
> Maybe, there is something I don't get, but why are the description of the energy part and the deflection part combined into one parametrization class. E.g. for pair production, the energy loss or asymmetry is calculated independent of the deflection (of course the latter depends on the energies sampled before) in two separated parametrizations.

The energy asymmetry and the deflection calculations are combined into one class because they are both needed for the calculation of the secondary particles. This process is (by design) calculated using one individual class that is responsible for calculating all features of the secondary particles, including the deflection, energy distribution among the particles and other features.

For pair production, they are also combined in one parametrization: `secondaries::KelnerKokoulinPetrukhinEpairProduction`, which calculates the asymmetry between the electron and the positron using `KelnerKokoulinPetrukhin`, and assumes no deflection of the secondary particles. There is of course a `crosssection::KelnerKokoulinPetrukhinEpairProduction` which is responsible for calculating `v`, but this is also the case for PhotoPairProduction (although the calculation of `v` is trivial here, since it is always 1).



> So why don't we split the PhotoPairProduction Secondary parametrization into one for the energy asymmetry and one for the angles? Thereby, the number of parametrizations can be kept to a minimum with no need to implement all combinations.

I tried to do this as good as possible with the current design by designing `PhotoPairProductionInterpolant` as a template class. If we want to calculate the energy asymmetry using a different crosssection (that is differential in rho), we can just create another instance of this template class. This is only a few lines of code if we already have implemented the underlying crosssection parametrization, see [here](https://github.com/tudo-astroparticlephysics/PROPOSAL/pull/247/files#diff-cc0e35301e4c23ad4207e899ced9d4540d7058f583cde4a248d56a36da52d3e2R8-R17).

sudojan:
> The energy asymmetry and the deflection calculations are combined into one class because they are both needed for the calculation of the secondary particles. This process is (by design) calculated using one individual class that is responsible for calculating all features of the secondary particles, including the deflection, energy distribution among the particles and other features.

I was thinking about the photonuclear parametrization, where e.g. the ALLM parametrization depends on a shadowing parametrization.

sudojan (approved):

PR #256, opened 2022-01-28
Catch negative values of the differential crosssection for PhotoQ2Intgration

Jean1995:
When using specific photonuclear crosssections with tau leptons, the differential crosssection becomes negative, which happend at low energies close to the tau mass. This causes dNdx to be negative as well. CubicInterpolation seems to have problems with these negative values, and crashes when trying to initialize the tables.

For example, this is the differential crosssection for taus in ice, using an EnergyCut of (Inf, 0.05), the ButkevichMikheyev parametrization and DuttaRenoSarvecivSeckel for the shadowing, at an energy of 1.1 tau masses:

This PR simply catches negative values of the differential crosssection and sets them to zero.


sudojan:
Does this only occur with the Parametrizations integrated over Q^2, or also for e.g. for Zeus?

sudojan (approved):

Jean1995:
> Does this only occur with the Parametrizations integrated over Q^2, or also for e.g. for Zeus?

I did not observe the issue for the RealPhotonAssumption cross sections - all crosssection objects could be created, and looking at some differential crosssections, they seem to be all-positive.

PR #260, opened 2022-02-09
Fix extrapolation to lower energies for PhotoproductionRhode

Jean1995:
The cross section values of PhotoproductionRhode are provided as tables values which are interpolated. The lowest energy for which an interpolation point is given is at 0.1 GeV. For energies below, until now, the cross section has just been extrapolated, leading to unphysical results. This can be seen in this plot by @asandrock:

This PR now ensures that we do not extrapolate the Rhode parametrization below our lowest point at 0.1 GeV.

However, this does not fix an issue with other photoproduction parametrizations.
Right now, their dNdx values look like this:

photoproduction_cross

We see that there is still an extrapolation for all other cross sections, which lead to an unphysical increase of the cross section towards lower energies. How should we behave here?
We could set the cross section to zero below a fixed energy, such as `M_PI**2  / 2*M_N`, but this would introduce a hard cut-off at this energy. However, this hard cut-off wouldn't be a big problem, because at energies of the cut off, the cross section is dominated by other processes.
The following plot indicated where this cut_off would be in this specific case:

photoproduction_cross

### DIFF
--- a/src/PROPOSAL/PROPOSAL/crosssection/parametrization/Photoproduction.h
+++ b/src/PROPOSAL/PROPOSAL/crosssection/parametrization/Photoproduction.h
@@ -36,6 +36,8 @@ namespace PROPOSAL {
 
             double GetLowerEnergyLim(const ParticleDef&, const Medium&, cut_ptr) const override;
 
+            double GetCutOff(const Component& comp) const;
+
             size_t GetHash(const ParticleDef&, const Medium& m, cut_ptr) const noexcept override;
 
             InteractionType GetInteractionType() const noexcept override;
--- a/src/PROPOSAL/PROPOSAL/crosssection/parametrization/Photoproduction.h
+++ b/src/PROPOSAL/PROPOSAL/crosssection/parametrization/Photoproduction.h
@@ -34,9 +34,7 @@ namespace PROPOSAL {
                 return 1.; // all energy is always lost, i.e. v=1
             };
 
-            double GetLowerEnergyLim(const ParticleDef&, const Medium&, cut_ptr) const override {
-                return 2. * ME; // TODO: other limit?
-            };
+            double GetLowerEnergyLim(const ParticleDef&, const Medium&, cut_ptr) const override;
 
             size_t GetHash(const ParticleDef&, const Medium& m, cut_ptr) const noexcept override;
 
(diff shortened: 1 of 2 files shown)


Jean1995:
I've now added a hard cut off. This is motivated by the energy necessary to create a pion.
The photoproduction crosssections now look like this:

photoproduction_cross

src/PROPOSAL/PROPOSAL/crosssection/parametrization/Photoproduction.h
```diff
@@ -34,9 +36,7 @@ namespace PROPOSAL {
                 return 1.; // all energy is always lost, i.e. v=1
             };
 
-            double GetLowerEnergyLim(const ParticleDef&, const Medium&, cut_ptr) const override {
-                return 2. * ME; // TODO: other limit?
```
sudojan, line 38: Does someone know, where the 2 m_e derives from?

src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/Photoproduction.cxx
```diff
@@ -16,6 +16,18 @@ size_t crosssection::Photoproduction::GetHash(const ParticleDef&, const Medium &
...
+double crosssection::Photoproduction::GetLowerEnergyLim(const ParticleDef&, const Medium& medium, cut_ptr) const {
+    double min_cut_off = INF;
+    for (const auto& c : medium.GetComponents())
+        min_cut_off = std::min(min_cut_off, GetCutOff(c));
+    return min_cut_off;
+};
+
+double crosssection::Photoproduction::GetCutOff(const Component& comp) const {
+    auto m_N = comp.GetAverageNucleonWeight();
+    return MPI + MPI * MPI / (2. * m_N);
```
sudojan, line 28: Maybe it is better to show, where this comes from, which is `crosssection::Photonuclear::GetKinematicLimits`

src/PROPOSAL/PROPOSAL/crosssection/parametrization/Photoproduction.h
Jean1995: From the cutoff that we used for photons, because this is the energy below which no new electron-positron pairs can be produced.

src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/Photoproduction.cxx
sudojan: at least
initially, I thought it can be defined in one place and then used in multiple applications, but maybe the overhead is too large overhead

sudojan (approved):

PR #261, opened 2022-02-16
Fix windows CI

Jean1995:
Github actions recently updated the default Windows runner from Windows Server 2019 to Windows Server 2022.
It looks like this not yet completely compatible with conan, at least they don't provide prebuilt packages for Visual Studio 17 which comes with WS2022.

Therefore, I would suggest that we keep testing on Windows Server 2019 until conan figured out all possible issues and provides the binaries.


sudojan (approved):
I created Issue #262 so that we don't forget this temporary fix.

PR #265, opened 2022-02-22
Add γ->µµ as a new interaction process

Jean1995:
This PR adds photomupairproduction, i.e. the process γ->µµ, as a new process to PROPOSAL.

It includes a parametrization based on the bremsstrahlung crosssection for muons by Kelner, Kokoulin and Petrukhin, which has been modified by @asandrock for pair production by photons. 

There is also a secondary production calculator included, where the energy distribution is sampled using the above-mentioned parametrization, while the muon angles are calculated using the Θ=MMU/E approximation.

### DIFF
--- a/src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/PhotoMuPairProduction.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/PhotoMuPairProduction.cxx
@@ -51,7 +51,7 @@ double crosssection::PhotoMuPairSandrock::DifferentialCrossSection(
     // calculate elastic formfactor Phi
     aux1 = MMU / ME * B * Z3 / (1. + B * Z3 * SQRTE * delta / ME);
     aux2 = D / (1. + delta * (D * SQRTE - 2.) / MMU);
-    if (Z <= 1)
+    if (Z > 1)
         aux2 *= (1. - 1. / Z); // influence of the nucleus excitation
     auto Phi = std::log(aux1) - std::log(aux2);
 
--- a/src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/PhotoMuPairProduction.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/PhotoMuPairProduction.cxx
@@ -1,0 +1,68 @@
+
+#include <cmath>
+
+#include "PROPOSAL/crosssection/parametrization/PhotoMuPairProduction.h"
+#include "PROPOSAL/Constants.h"
+#include "PROPOSAL/medium/Components.h"
+#include "PROPOSAL/particle/Particle.h"
+
+using namespace PROPOSAL;
+using std::make_tuple;
+
+double crosssection::PhotoMuPairProduction::GetLowerEnergyLim(
+        const ParticleDef&) const noexcept {
+    return 2. * MMU * SQRTE;
+}
+
+crosssection::KinematicLimits crosssection::PhotoMuPairProduction::GetKinematicLimits(
+        const ParticleDef&, const Component&, double energy) const noexcept {
+    // x is the integration variable here
+    KinematicLimits lim{};
+
+    if (energy <= 2. * MMU * SQRTE) {
+        lim.v_min = 0.5;
+        lim.v_max = 0.5;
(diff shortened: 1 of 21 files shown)


src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/PhotoMuPairProduction.cxx
```diff
@@ -0,0 +1,71 @@
...
+    return std::make_unique<param_t>(*this);
+}
+
+double crosssection::PhotoMuPairSandrock::DifferentialCrossSection(
+        const ParticleDef&, const Component& comp, double energy, double x) const
+{
+    double aux1, aux2;
+
+    auto delta = MMU * MMU / (2. * energy * x * (1. - x));
+    auto B = 183.;
```
sudojan, line 45: The radiation logarithm depends on the medium and should be treated like in the original bremsstrahlung cross section, this parametrization derives from.
The same holds for the inelastic radiation logarithm Bprime.

Jean1995: You mean
```suggestion
    auto B = comp.GetLogConstant();
```
and for Bprime we use `comp.GetBPrime()`?

Jean1995:
I've made some plots to validate the implementation and show the effects of it.

To test the sampling of `x` (i.e. the fraction of energy going into one of the muons), I've sampled 1e5 losses at an initial energy E_i and compare the results to the differential cross section:

x_sampling

This also looks comparable to the results shown in FZKA7525, Figure 2:

image

I've also looked at the sampling of all photon interactions that are now possible. I sampled 5e5 interactions for different energies, and compared their frequencies to the ratios of the cross sections:

interaction_sampling

src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/PhotoMuPairProduction.cxx
asandrock: Yes, that would be more exact. 183 is the value for a Thomas-Fermi atom (approximating the electrons as an ideal Fermi gas; roughly OK for high Z), while the values from comp.GetLogConstant() are determined from a calculation using the Hartree-Fock model (self-consistent iterative determination of atomic electron wavefunctions). The change is not large, for nitrogen B is 176.6 in the Hartree-Fock method, resulting in a change of about 0.2% in the complete screening regime compared to the Thomas-Fermi value; but this way it becomes consistent to the KKP muon bremsstrahlung cross section.

sudojan:
This looks all very good to me.
The remaining tasks/questions, I can think of, would be:
- Describe the process in the documentation, like [here](https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/master/docs/config_docu.md)
- Should this process be included in the default cross-sections for photons?
- Should the advanced secondary calculation from FZKA7525 be included in this PR (as a second parametrization)?
- What should be the name of the energy loss parametrization and the name for the secondary parametrization? The energy loss calculation is mainly taken from the KelnerKokoulinPetrukhin Parametrization from the Bremsstrahlung with modifications by Sandrock (to be published).

Jean1995:
> This looks all very good to me. The remaining tasks/questions, I can think of, would be:
> 
>     * [x]  Describe the process in the documentation, like [here](https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/master/docs/config_docu.md)
> 
>     * [x]  Should this process be included in the default cross-sections for photons?
> 
>     * [x]  Should the advanced secondary calculation from FZKA7525 be included in this PR (as a second parametrization)?
> 
>     * [ ]  What should be the name of the energy loss parametrization and the name for the secondary parametrization? The energy loss calculation is mainly taken from the KelnerKokoulinPetrukhin Parametrization from the Bremsstrahlung with modifications by Sandrock (to be published).

The documentation has been updated. I've also added a comment in the source file about the source of the cross section.

We decided not to include PhotoMuPairProduction in the default cross sections, and we decided to not (yet) include the parametrization from FZKA7525.

The remaining question is the naming. For now, the parametrizations (for both cross section and secondaries) are just called 'Sandrock'. Should we keep this?

sudojan (approved):

Issue #246, opened 2022-01-15
Scattering problem at borders


Jean1995:
When trying to propagate low-energy muons, we encountered some problems with multiple scattering at border transitions.

Firstly, I will try to explain the concept that is current used to treat multiple scattering at border transitions (i.e. transitions from one medium to another), and then I will go into more detail about what is going wrong:

### Concept

![drawing1](https://user-images.githubusercontent.com/15159319/149620441-e0567e23-5c2f-486c-bdca-98d74ea22d42.png)

In this sketch, the black circle segment indicated a medium/geometry transition. PROPOSAL tries to propagate a particle with a straight line to a next interaction (arrow 1). Now, multiple scattering is applied, and a deflected direction is sampled (arrow 2). PROPOSAL now checks whether this proposed step is valid: We detect that we cross a border, so this step is not valid! 

![drawing2](https://user-images.githubusercontent.com/15159319/149620526-016405a1-be1d-4098-b3de-8ba68465474c.png)

We have to sample the step again. This time we don't try to propagate the particle in a straight line to the next interaction, but we propagate it for a shorter distance (the distance that arrow 2 had until the border), which is indicated by arrow 3. Now, we apply multiple scattering on this straight line, and get arrow 4. Notice that the deflection angle of arrow 4 is smaller compared to the deflection angle that we previously sampled for arrow 2, since we propagate a smaller distance now.
Now, we check whether this proposed step (arrow 4) is valid. This time, the step is not valid because we do not reach the border: We are missing a distance that is indicated by the red dotted line!

![drawing3](https://user-images.githubusercontent.com/15159319/149620693-e2013244-c94a-475b-b53c-fbf5d239804b.png)

Finally, we get arrow 5, which is a bit longer than arrow 3. We apply multiple scattering on this proposed step and land at arrow 6. This now reaches the border (or rather: reaches the border within a geometrical uncertainty), so we accept this propagation step!

For high energy muons, this concept usually converges within 2 or 3 steps. We use this algorithm so that we do not under- or overestimate multiple scattering when we have to decrease the originally intended step length to to a border transition.
For low energy muons however, I realized that on the current master, this algorithm can take up to several hundreds or thousands steps to converge! 
[...]

PR #267, opened 2022-03-04
Rewrite Propagator::AdvanceParticle method

Jean1995:
## Problem description

The `Propagator::AdvanceParticle` method is responsible for doing the continuous steps of the particle propagation. This includes the calculation of the continuous energy losses as well at the particle displacement, where we also have to correctly consider the effects of multiple scattering and take into account border transitions.

However, the border transitions (in combination with multiple scattering) have been the focus of several issues in the past. One current issue that has to do with scattering and border transitions has been described in detail in issue #246.

Another issue on the master branch occurred when I tried to solve this issue.
Let's assume we have a particle that propagated to a sector border in the previous step. During the next continuous step, it can be possible that the particle immediately scatters back into the old sector ("backscattering"). This situation is shown in the following sketch.

backscatter_small

This behavior is not treated correctly on the current master branch, as PROPOSAL fails to realize that the particle moved back to the old sector during the transition. This could lead to runtime errors such as:

```
RuntimeError: Root must be bracketed in Bisection method!
```

## Fix

I started to fix both issues, which are originated in `Propagator::AdvanceParticle`. However, I did not only realize that fixing the code wasn't straight-forward, I also realized that the old code has been rather hard to understand.
Therefore, I decided to completely rewrite the code - Although the basic idea of the algorithm is still the same.

For border transitions, the idea is still to iterate multiple scattering steps until our combination of scattering angle and propagation distance reaches the sector transition (This concept is described in more detail in #246).

The fix for issue #246 is to keep the same random numbers for all iterations. If we run into the "loop" problem described in #246, we discard the current set of random numbers. This leads to a minimal directional bias for specific border transitions. However, this bias only occurs for very specific geometrical constellations, and is also minimal. Furthermore, this bias has basically been included in the old algorithm as well, and could even lead to very significant biases (an example will be given below).

## Runtime comparisons and check for biases

As described in #246, the algorithm has to perform way too many iterations since the random numbers are discarded for every iteration. Therefore, this fix is expected to have a significant impact on the runtime, at least for low-energy muons as well as propagations where border crossings are performed.

To check this, I created a propagation environment where I simply propagated muons until their decay, and put a border transition into the propagation environment at 0.25 times of the mean free path lenght of the muon.

If we compare the runtimes of this PR with the master, we see that the new algorithm improves the runtime per propagation by almost a factor of two for low energies. As we go to higher energies, the difference becomes negligible, which is expected since the issue in the algorithm becomes less severe for higher energies, and since the border transition is not as important runtime-wise for higher energies.

runtime_comparison

Furthermore, we can look at the deflection from the shower axis at the end of the propagation.
If we compare the results from the old master with the PR, we see that particles have a higher overall deflection from the shower axis now. This is also to be expected: In the old algorithm, we basically sampled new deflections in each iteration, until two consecutive deflections pointed into the same direction by chance. However, the probability to be similar is much smaller for smaller deflections than for higher deflections (since the solid angle is smaller for smaller deflections.) This lead to a bias towards smaller deflections in the old algorithm.

deflection_comparison_1228 113526588618

## Another bias check

In this check, we propagate particles in parallel and very close to a sector border (1‰ of the mean free path length away from the sector border). Note that both sectors have the same medium. If we look at the final particle positions with this PR, the histogram looks like this:

deflection_bias_1228 113526588618

We see a small bias where the particles are rather scattering away from the sector border (at around 0cm) instead of towards the sector border. This bias is expected, since in some cases, the random numbers are discarded.

If we look at the same plot created on the master branch, we see that the bias is actually much more significant:

deflection_bias_1228 113526588618_master

This is because in the old algorithm, directions without a sector transition have always been preferred. Note that this is a constructed example where we propagate a particle in perfect parallel to a sector transition. It should also be noted that this propagation on the master has been very unstable and lead to a lot of runtime errors. The reason for this is the "backscattering" issue. This has been fixed with this PR

## Just some nice plots

To visualize the results, I also created some plots with medium transitions (between Hydrogen and Stone) which I didn't want to withhold. They basically just show visually that nothing crazy is happening at sector transitions.

<details>
<summary>Click to expand</summary>

example1

example2

example3

example4

</details>

## Conclusion

This PR has the following advantages:

- Simpler code in `AdvanceParticle`
- Better runtime at border transitions, especially for small energies (fix issue #246)
- Minimizing the bias at border transitions
- Fixing runtime errors for "backscattering"

### DIFF
--- a/src/PROPOSAL/detail/PROPOSAL/Propagator.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/Propagator.cxx
@@ -173,9 +173,11 @@ int Propagator::AdvanceParticle(ParticleState &state,
         return random_numbers[i++%4];
     };
 
+    // Iterate combinations of step lengths and scattering angles until we have
+    // reached an interaction, a sector border or the maximal propagation distance
     do {
         // Calculate grammage, energy and distance for step
-        if (energy != -1 && grammage == -1 && distance == -1) {
+        if (energy != -1 && distance == -1) {
             // Calculate grammage and distance from given energy
             grammage = utility.LengthContinuous(state.energy, energy);
             try {
@@ -183,7 +185,7 @@ int Propagator::AdvanceParticle(ParticleState &state,
             } catch (const DensityException&) {
                 distance = INF;
             }
-        } else if (energy == -1 && grammage == -1 && distance != -1) {
+        } else if (energy == -1 && distance != -1) {
             // Calculate energy and grammage from given distance
             auto grammage_step = density->Calculate(state.position, state.direction, distance);
             if (grammage_step < grammage_next_interaction) {
@@ -195,14 +197,21 @@ int Propagator::AdvanceParticle(ParticleState &state,
                 for (auto& r: random_numbers) {
                     r = rnd_generator();
                 }
-                Logging::Get("proposal.propagator")->debug("Discard current set of random numbers for multiple"
-                                                           " scattering since we are stuck in a loop.");
+                Logging::Get("proposal.propagator")->debug("Unable to find a valid combination of propagation step "
+                                                           "length and multiple scattering angle for this set of "
+                                                           "random numbers. Resample set of random numbers.");
                 grammage = grammage_next_interaction;
                 energy = energy_next_interaction;
-                distance = density->Correct(state.position, state.direction, grammage, max_distance);
+                try {
+                    distance = density->Correct(state.position, state.direction, grammage, max_distance);
+                } catch (const DensityException&) {
+                    distance = INF;
+                }
             }
         } else {
-            throw std::logic_error("Error in AdvanceParticle: We shouldn't end up here!");
+            throw std::logic_error("Error in AdvanceParticle: Either both distance and final energy for the next "
+                                   "iteration step are known, or neither of them are known. This should never happen "
+                                   "and would indicate an algorithmic error!");
         }
 
         // Calculate scattering proposal
--- a/src/PROPOSAL/detail/PROPOSAL/Propagator.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/Propagator.cxx
@@ -86,6 +86,9 @@ Secondaries Propagator::Propagate(const ParticleState& initial_particle,
 
         advancement_type = AdvanceParticle(state, energy_at_next_interaction,
             max_distance, rnd, current_sector);
+        utility = get<UTILITY>(current_sector);
+        density = get<DENSITY_DISTR>(current_sector);
+
         track.push_back(state, InteractionType::ContinuousEnergyLoss);
 
         switch (advancement_type) {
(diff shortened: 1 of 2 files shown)


src/PROPOSAL/detail/PROPOSAL/Propagator.cxx
```diff
@@ -139,93 +142,112 @@ Interaction::Loss Propagator::DoStochasticInteraction(ParticleState& p_cond,
...
+                grammage = grammage_step;
+                energy = utility.EnergyDistance(state.energy, grammage);
+            } else {
+                // we are unable to reach `distance` before we reach the next interaction
+                // this means we are stuck in a loop, and need to discard the current set of random numbers
+                for (auto& r: random_numbers) {
+                    r = rnd_generator();
+                }
+                Logging::Get("proposal.propagator")->debug("Discard current set of random numbers for multiple"
+                                                           " scattering since we are stuck in a loop.");
```
sudojan, line 199: This can be more descriptive. Explain what the loop is.

src/PROPOSAL/detail/PROPOSAL/Propagator.cxx
```diff
@@ -139,93 +142,112 @@ Interaction::Loss Propagator::DoStochasticInteraction(ParticleState& p_cond,
...
+                grammage = grammage_next_interaction;
+                energy = energy_next_interaction;
+                distance = density->Correct(state.position, state.direction, grammage, max_distance);
```
sudojan, line 202: Cannot this raise a DensityException like a couple of lines above?

src/PROPOSAL/detail/PROPOSAL/Propagator.cxx
```diff
@@ -139,93 +142,112 @@ Interaction::Loss Propagator::DoStochasticInteraction(ParticleState& p_cond,
...
             }
+        } else {
+            throw std::logic_error("Error in AdvanceParticle: We shouldn't end up here!");
```
sudojan, line 205: Better error message

src/PROPOSAL/detail/PROPOSAL/Propagator.cxx
```diff
@@ -139,93 +142,112 @@ Interaction::Loss Propagator::DoStochasticInteraction(ParticleState& p_cond,
...
+            // reached interaction
+            advancement_type = ReachedInteraction;
+        } else if (distance <= distance_to_border && distance == max_distance) {
+            // reached max distance
+            advancement_type = ReachedMaxDistance;
+        } else if (std::abs(distance - distance_to_border) <= PARTICLE_POSITION_RESOLUTION) {
+            // reached geometry border
+            advancement_type = ReachedBorder;
+            distance = distance_to_border;
+        } else {
```
sudojan, line 237: When do we run into this case, only if the distance is set to INF due to a DensityError?
Or what other cases exist?

src/PROPOSAL/detail/PROPOSAL/Propagator.cxx
```diff
@@ -139,93 +142,112 @@ Interaction::Loss Propagator::DoStochasticInteraction(ParticleState& p_cond,
...
+            energy = -1;
+            grammage = -1;
+        }
+    } while (advancement_type == InvalidStep);
 
-    state.time = state.time
-        + utility.TimeElapsed(state.energy, E_f, advance_grammage,
-            density->Evaluate(state.position));
-    state.position = new_position;
+    state.time = state.time + utility.TimeElapsed(state.energy, energy, grammage, density->Evaluate(state.position)); // TODO: should the energy passed here be the randomized energy or not?
```
sudojan, line 246: This is quite a good point. Historically, the randomized energy has always been calculated after the propagation time.
The reason for this was, that the time and the distance were calculated together in the old `AdvanceParticle` method, see e.g. [here](https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/v3.0.0/private/PROPOSAL/Propagator.cxx#L756).
Since then we have kept the order of calculation.
Thinking of it, the time calculation should depend on the randomized energy.
But this should be discussed in the next PROPOSAL meeting.

sudojan (changes requested):
The code requests are inline.
Further requests regarding the plots.
- Add errors to the runtime plots.
- Use a binning centered around zero for the bias histogram.

Something, which has not necessarily to be addressed in this issue:
- Can we quantify the bias due to the resampling, depending on the distance to the border (in grammage) and the angle?

src/PROPOSAL/detail/PROPOSAL/Propagator.cxx
```diff
@@ -139,93 +142,112 @@ Interaction::Loss Propagator::DoStochasticInteraction(ParticleState& p_cond,
...
+    int i = 0;
+    auto rnd = [&rnd_generator, &i, &random_numbers]() {
+        if (random_numbers[i%4] == -1)
+            random_numbers[i%4] = rnd_generator();
+        return random_numbers[i++%4];
+    };
+
+    do {
+        // Calculate grammage, energy and distance for step
+        if (energy != -1 && grammage == -1 && distance == -1) {
```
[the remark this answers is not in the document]
Jean1995: Code relict from when it was not clear what cases are actually existing.
The -1 indicates that a variable is not known. But since this is always the case for grammage, we can delete it.

src/PROPOSAL/detail/PROPOSAL/Propagator.cxx
Jean1995: I've added a comment of the loop to describe shortly what it does.
And I also added a bit more context to describe why the random numbers are discarded in this case.

src/PROPOSAL/detail/PROPOSAL/Propagator.cxx
Jean1995: You mean the last else bracket?

As indicated by the comment, this means that we didn't finish our iteration: With the current combination of our propagation step length and the scattering angle, we do not hit the sector border (we either overshoot the border, or we undershoot it). Therefore we have to perform another iteration step until we have reached the sector border within our particle precision (... or until we reached the maximum propagation distance, or reached an interaction, which are the other cases that can appear)

Jean1995:
>     * Use a binning centered around zero for the bias histogram.

On master, the bias looks like this:

<p float="left">
  
   
</p>

With the PR, it looks like this:

<p float="left">
  
   
</p>

Jean1995:
>     * Add errors to the runtime plots.

These are some new plots where I plot the std of the runtime as an error band.

runtime_comparison

Note that the fluctuations are pretty high, which means that the mean values of the runtime are also fluctuating a lot (it looks slightly different every time I re-create the plots). But the runtime improvement at low energies is always visible.

src/PROPOSAL/detail/PROPOSAL/Propagator.cxx
Jean1995: If we decide to change the order here, I think we should add this in a separate PR. In my opinion, this is pretty significant, and should not be "hidden" in this restructuring PR.

So I would say that we merge it with the "old" order, and make a new PR where we only change the order of the continuous randomization calculation.

sudojan: ok

sudojan (approved):

PR #269, opened 2022-03-11
Add UnitTests for PhotoMuPairProduction

Jean1995:
Add some UnitTests for PhotoMuPairProduction.
We check dNdx and the dNdx interpolant, we check that the interaction is only-stochastic (i.e. no dEdx), we check that losses are always catastrophic (i.e. v=1) and we check the x sampling.

While writing the tests, I also realized that the PhotoMuPairProduction parametrization has not yet been registered as "only_stochastic", so this has been fixed.

### DIFF
--- a/tests/PhotoMuPair_TEST.cxx
+++ b/tests/PhotoMuPair_TEST.cxx
@@ -1,0 +1,139 @@
+#include "gtest/gtest.h"
+#include <nlohmann/json.hpp>
+
+#include "PROPOSAL/crosssection/Factories/PhotoMuPairProductionFactory.h"
+#include "PROPOSAL/secondaries/parametrization/photomupairproduction/PhotoMuPairProductionBurkhardtKelnerKokoulin.h"
+#include "PROPOSAL/medium/MediumFactory.h"
+#include "PROPOSAL/particle/ParticleDef.h"
+#include "PROPOSALTestUtilities/TestFilesHandling.h"
+
+using namespace PROPOSAL;
+
+ParticleDef getParticleDef(const std::string& name)
+{
+    if (name == "Gamma") {
+        return GammaDef();
+    } else {
+        EXPECT_TRUE(false);
+        return MuMinusDef();
+    }
+}
+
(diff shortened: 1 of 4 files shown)


tests/PhotoMuPair_TEST.cxx
```diff
@@ -0,0 +1,139 @@
...
+        ParticleDef particle_def = getParticleDef(particleName);
+        auto medium = CreateMedium(mediumName);
+
+        nlohmann::json config;
+        config["parametrization"] = parametrization;
+
+        auto cross = make_photomupairproduction(particle_def, *medium, true, config);
+
+        dNdx_new = cross->CalculatedNdx(energy);
+        EXPECT_NEAR(dNdx_new, dNdx_stored, 1e-3 * dNdx_stored);
```
sudojan, line 87: Can the accuracy be increased to 1e-6 or is this the limit?

sudojan, line 87: Can the accuracy be increased to 1e-6 or is this the limit?

tests/PhotoMuPair_TEST.cxx
```diff
@@ -0,0 +1,139 @@
...
+        ParticleDef particle_def = getParticleDef(particleName);
+        auto medium = CreateMedium(mediumName);
+
+        nlohmann::json config;
+        config["parametrization"] = parametrization;
+
+        auto cross = make_photomupairproduction(particle_def, *medium, false, config);
+
+        dNdx_new = cross->CalculatedNdx(energy);
+        EXPECT_NEAR(dNdx_new, dNdx_stored, 1e-5 * dNdx_stored);
```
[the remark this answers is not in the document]
Jean1995: Done

tests/PhotoMuPair_TEST.cxx
Jean1995: No, the interpolation precision is actually limited with the current settings since the interpolation is not "trivial" for all energies.

For example for 1e3 to 1e4 MeV:

image

sudojan (approved):

Issue #272, opened 2022-03-28
Impementation details of Continuous Randomization in Propagator


Jean1995:
We already had some discussions about the current implementation of Continuous Randomization inside the `Propagator::Propagate` method [in this PR](https://github.com/tudo-astroparticlephysics/PROPOSAL/pull/267#discussion_r820256036). 

However, there is another point that we might want to change. I learned about this after some discussion with a PROPOSAL user:

The user propagated muons to their rest mass (i.e. he used `Propagator::Propagate(init_state)`). He wondered why the particles did not read their rest mass, where I explained that this might be due to muons decaying before they reached their rest mass. We disabled the muon decay by setting the muon lifetime to infinte, however, there were still a lot of muons that did not reach their rest mass. We found out that this is due to the continuous randomization.

The final energies of (stable) muons, when continuous randomization is enabled, looks like this:

![contrand](https://user-images.githubusercontent.com/15159319/160411437-a9398afd-85f3-4bc4-a7d8-cc9a4d019403.png)

If we disable continuous randomization, as expected, all muons reached their rest mass at the end of the propagation.

What is happening here is that when we have our last continuous propagation step, i.e. the step with `E_f = m_mu`, `Propagator::Propagate` randomizes the final energy before the propagation is stopped.

The same effect can be seen if we set a minimal propagation energy, for example of 1000 MeV:

![contrand_1000](https://user-images.githubusercontent.com/15159319/160411894-e8c2c8f0-5490-473c-9a4c-7d68d07f9560.png)

I feel like this should not be the case: If we tell the Propagator that we want to propagate the particle to a certain energy (or to its rest mass), the final particle energy should be equal to this value whenever possible (exceptions are stochastic losses, which may cause the final energy to be below the fixed value, or fatal interaction such as a particle decay). 

In my opinion, the energy should not be randomized for this last step in case we are propagating to a specific energy. However, in this case, the last step would not be randomized at all. I am not sure whether this is a problem for us.

PR #276, opened 2022-04-04
Avoid continuous randomization below min_energy

Jean1995:
This bug fixes the issue described in #272 

Continuous randomization did not take the `min_energy` option of the propagation into account.
This means that it was possible that a continuous step with a `final_energy` equal to `min_energy` ended with an energy below or above `min_energy`. In addition, it was possible that the final energy of a continuous step could be randomized to an energy below `min_energy`.

To fix this, there are two changes to the Propagation routine:

1. If our continuous step ends at `min_energy`, do not apply continuous randomization
2. If we do continuous randomization, don't randomize the final energy to an energy below `min_energy`. For this a `min_energy` parameter has been introduced to all `EnergyRandomize` functions. 

These are the results:

A propagation of stable muons with a `min_energy` of 1000 MeV looked like this on the master:

master_with_contrand

With this PR, it looks like this:

pr_with_contrand

There are still particles that end with an energy below 1000 MeV, however this is caused by stochastic losses where the remaining energy becomes smaller than `min_energy`. This is unavoidable.

We can also compare this to the energy histogram of the same propagation, but with continuous randomization disabled:

pr_without_contrand

The result is identical to the histogram with continuous randomization enabled, which means that we don't see any artifacts from the continuous randomization anymore!

Another issue were muons propagated until their rest mass. On the master, the energy histogram looked like this:

contrand

With this PR, all muons end with their rest mass, although continuous randomization is enabled:

pr_with_contrand_rest_mass


src/PROPOSAL/PROPOSAL/propagation_utility/ContRandBuilder.h
```diff
@@ -27,11 +27,11 @@ template <class T> class ContRandBuilder : public ContRand {
     }
 
     inline double EnergyRandomize(
-        double initial_energy, double final_energy, double rnd) final
+        double initial_energy, double final_energy, double rnd, double min_energy = 0) final
     {
         auto std = std::sqrt(Variance(initial_energy, final_energy));
         return SampleFromGaussian(final_energy, std, rnd,
-                                  std::min(disp->GetLowerLim(), final_energy),
+                                  std::max(disp->GetLowerLim(), min_energy),
```
sudojan, line 34: So this is the key change of this PR.
Before, the minimum of the lower limit and the final energy was taken. Now, the final energy is not considered anymore. I don't fully understand why this was done in the past. But if there was a good reason for this, maybe an edge case, we must check if this edge case is still treated correctly.

Jean1995:
I'll add some UnitTests to check that this behavior doesn't appear anymore.

src/PROPOSAL/PROPOSAL/propagation_utility/ContRandBuilder.h
sudojan: and if the sampled final energy is below the lower limit of interpolated displacement energies (which is usually the mass of the particle or a very small energy for massless particles like photons or neutrinos), this raises other issues, I think.
Just wanted to be sure.
For me, the current change makes more sense.
However, maybe a comment could help to solve an issue in the future, in case we overlook something, here.

I would enhance the `Constraints` of the `ContinuousRandomization` UnitTest with a minimal energy. Those events, ending with a stochastic interaction instead of a continuous loss, should of course be avoided as they can be below the threshold, as explained above.

sudojan (approved):

src/PROPOSAL/PROPOSAL/Propagator.h
```diff
@@ -33,9 +33,8 @@ class Propagator {
 private:
     Interaction::Loss DoStochasticInteraction(
         ParticleState&, PropagationUtility&, std::function<double()>);
-    int AdvanceParticle(ParticleState& p_cond, const double E_f,
-                        const double max_distance,
-        std::function<double()> rnd, Sector& current_sector);
+    int AdvanceParticle(ParticleState&, const double, const double,
```
maxnoe, line 36: Why are there no parameter names anymore?

Jean1995: No real reason. They are not needed in the header file. But I could put them back for convenience.

PR #277, opened 2022-04-05
Catch negative dNdx values

### LINKED ISSUE #273: Negative rates causing Runtime Errors
@pascalgutjahr tried to propagate 2e6 MeV muons in ice, using a `v_cut` of 0.001.
After some propagations, PROPOSAL threw the `no dNdx for this energy defined` runtime exception.

After some debugging, I learned that the problem is due to interpolation issues for the Ionization cross section. The dNdx interpolant looks like this:

interpolant

For E ≈ 170 MeV, the cross sections becomes negative due to overshooting. This is because the integrand becomes hard to interpolate for very small v, as for the 0.001 used in this case.

Our method which samples the stochastic loss, [`Interaction::SampleLoss`](https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/master/src/PROPOSAL/detail/PROPOSAL/propagation_utility/Interaction.cxx#L38), uses the sum of all rates to sample the interaction type, component and size of the loss. A negative rate therefore leads to wrong calculations here.

### Approach to solve this issue on branch [negative_rate_error](https://github.com/tudo-astroparticlephysics/PROPOSAL/compare/negative_rate_error)

The real problem here is the interpolation overshooting. However, we are probably unable to ensure that all interpolants for all possible combinations of EnergyCutSettings and media are going to perform correctly.

Therefore, to avoid these miscalculations, I am catching all negative rates calculated by the interpolant and set them to zero. This way, we "silently" ignore the interpolations problems and just ignore the corresponding interaction type.

However, for the case described above, this means that we are going to sample a bremsstrahlung interaction which should in reality be very unlikely:

interpolant

### Alternative approaches?

Alternatively, we could check for negative rates at some point (for example in [`Interaction::Rates`](https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/master/src/PROPOSAL/detail/PROPOSAL/propagation_utility/Interaction.cxx#L70)) and throw a warning (in addition to setting the rate to zero). We could also just throw a runtime error which stops the propagation.

If we show at least a warning, the user would see that something is wrong, probably with the interpolation. The user might want to change the InterpolationSettings to solve these issues.

If we even throw a runtime error, the problem is that PROPOSAL would be more likely to crash, even if the error only affects a negligible kinematic range (e.g. low muon energies as in the case described above). This way, we would _force_ the user to change their propagation or interpolation settings.

Jean1995:
Fixes #273 

This is basically a short-term fix to catch negative dNdx values.
We set the dNdx value to zero, but we also warn the user, so that he can fix the interpolation issues if necessary (e.g. by adapting the interpolation settings).

As a long term solution, as discussed in #273 and in the PROPOSAL developer meeting, we might want to think about 

1. ...implementing an automatic self-test for interpolation tables, which checks whether everything is consistent (for example by comparing interpolated and integrated values for the whole parameter range)
2. ...implementing a routine which automatically adjust the interpolating settings for a given set of parameters.

### DIFF
--- a/src/PROPOSAL/detail/PROPOSAL/crosssection/CrossSectionDNDX/CrossSectionDNDXInterpolant.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/crosssection/CrossSectionDNDX/CrossSectionDNDXInterpolant.cxx
@@ -71,7 +71,13 @@ double CrossSectionDNDXInterpolant::Calculate(double energy, double v)
         return 0.;
     auto lim = GetIntegrationLimits(energy);
     v = retransform_v(lim.min, lim.max, v);
-    return std::max(interpolant.evaluate(std::array<double, 2> { energy, v }), 0.);
+    auto dNdx = interpolant.evaluate(std::array<double, 2> { energy, v });
+    if (dNdx < 0) {
+        logger->warn("Negative dNdx value for E = {:.4f} MeV, v = {:.4f} "
+                     "detected. Setting dNdx to zero.", energy, v);
+        dNdx = 0.;
+    }
+    return dNdx;
 }
 
 double CrossSectionDNDXInterpolant::GetUpperLimit(double energy, double rate)
--- a/src/PROPOSAL/detail/PROPOSAL/crosssection/CrossSectionDNDX/CrossSectionDNDXInterpolant.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/crosssection/CrossSectionDNDX/CrossSectionDNDXInterpolant.cxx
@@ -71,7 +71,7 @@ double CrossSectionDNDXInterpolant::Calculate(double energy, double v)
         return 0.;
     auto lim = GetIntegrationLimits(energy);
     v = retransform_v(lim.min, lim.max, v);
-    return interpolant.evaluate(std::array<double, 2> { energy, v });
+    return std::max(interpolant.evaluate(std::array<double, 2> { energy, v }), 0.);
 }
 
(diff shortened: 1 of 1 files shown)


src/PROPOSAL/detail/PROPOSAL/crosssection/CrossSectionDNDX/CrossSectionDNDXInterpolant.cxx
```diff
@@ -71,7 +71,13 @@ double CrossSectionDNDXInterpolant::Calculate(double energy, double v)
         return 0.;
     auto lim = GetIntegrationLimits(energy);
     v = retransform_v(lim.min, lim.max, v);
-    return interpolant.evaluate(std::array<double, 2> { energy, v });
+    auto dNdx = interpolant.evaluate(std::array<double, 2> { energy, v });
+    if (dNdx < 0) {
+        logger->warn("Negative dNdx value for E = {:.4f} MeV, v = {:.4f} "
```
sudojan, line 76: Is it possible to inform the user also about the interaction type?

Jean1995: Yes.
The error messages now look like this for example:

```
[2022-04-06 10:16:05.890] [CrossSection.DNDX] [warning] Negative dNdx value for E = 228.6277 MeV, v = 1.0000 detected in interaction type Ioniz. Setting dNdx to zero.
```

sudojan (approved):

PR #278, opened 2022-04-08
Fix Secondaries calculation when Multiple Scattering is enabled

Jean1995:
Most of the calculations in `Secondaries` didn't take multiple scattering correctly into account.
This influenced most of the calculations, including the calculation of exit points, entry points and approach points.

When calculating the possible entry point of a particle, the directions of the individual track points have been used.
However, since multiple scattering also introduces a displacement of the particle, these directions are not identical to the direct path between two track points.

This leads to a situation like in this sketch: We want to calculate the entry point with a specific geometry (red circle). However, the algorithm used the red arrow, which is the direction at track point 1. What should be done, and which is now implemented, is to use the direct line between two track points (dotted line) to calculate intersections.
image

This hasn't been caught in the UnitTests because there, multiple scattering has been disabled as well. Therefore, I activated multiple scattering in the UnitTest. Furthermore, I added UnitTests for the `GetClosestApproachPoint` method, which hasn't been tested before.

### DIFF
--- a/src/PROPOSAL/PROPOSAL/Secondaries.h
+++ b/src/PROPOSAL/PROPOSAL/Secondaries.h
@@ -47,7 +47,7 @@ using Sector = std::tuple<std::shared_ptr<const Geometry>, PropagationUtility,
 class Secondaries {
 
 public:
-    Secondaries(std::shared_ptr<ParticleDef>, std::vector<Sector>);
+    Secondaries(std::shared_ptr<ParticleDef> p_def, std::vector<Sector> sectors);
 
     // Operational functions to fill and access track
     void reserve(size_t number_secondaries);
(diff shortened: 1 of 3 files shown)


src/PROPOSAL/detail/PROPOSAL/Secondaries.cxx
```diff
@@ -225,23 +236,30 @@ std::shared_ptr<ParticleState> Secondaries::GetExitPoint(
...
+        auto pos_i = track_.at(i).position;
+        auto pos_f = track_.at(i+1).position;
+
+        auto displacement = pos_f - pos_i;
+        auto dist_i_f = displacement.magnitude();
+        displacement.normalize();
+
+        auto distance_to_closest_approach
+            = geometry.DistanceToClosestApproach(pos_i, displacement);
+        if (distance_to_closest_approach == dist_i_f) {
```
sudojan, line 255: this if-statement should be as precise as the PARTICLE_POSITION_RESOLUTION, see 3 lines below

src/PROPOSAL/detail/PROPOSAL/Secondaries.cxx
```diff
@@ -225,23 +236,30 @@ std::shared_ptr<ParticleState> Secondaries::GetExitPoint(
...
-                return std::make_unique<ParticleState>(track_.front());
-            auto prev_pos = track_.at(i-1).position;
-            auto prev_dir = track_.at(i-1).direction;
-            auto displacement = geometry.DistanceToClosestApproach(prev_pos,
-                                                                   prev_dir);
-            auto closest_approach = RePropagateDistance(track_.at(i-1),
-                                                        displacement);
+
+            auto closest_approach = RePropagateDistance(
+                    track_.at(i), displacement, distance_to_closest_approach);
```
sudojan, line 262: Are these bounds checks, which are also always computed in `GetEntryPoint` and `GetExitPoint`, really necessary? Maybe using `[]` instead of `.at()` would increase the performance.

src/PROPOSAL/PROPOSAL/Secondaries.h
```diff
@@ -101,8 +101,10 @@ class Secondaries {
     std::vector<ContinuousLoss> GetContinuousLosses(const Geometry&) const;
 
 private:
-    ParticleState RePropagateDistance(const ParticleState&, double) const;
-    ParticleState RePropagateEnergy(const ParticleState&, double, double) const;
+    ParticleState RePropagateDistance(const ParticleState&, const Cartesian3D&,
```
sudojan, line 104: please add the parameter names, although this is not necessary, it improves the code readability; see the discussion [here](https://github.com/tudo-astroparticlephysics/PROPOSAL/pull/276#pullrequestreview-931656503)

src/PROPOSAL/detail/PROPOSAL/Secondaries.cxx
Jean1995: Since it is safe that the index exist inside the loop (as long at the header is correct), we should be fine to use `[]`

sudojan (approved):

PR #280, opened 2022-04-20
error in bremsstrahlung deflection for low energies 

pascalgutjahr:
The parametrizations of Van Ginneken are check for energies `3 GeV < E < 30 TeV`. For high energy losses `nu > 0.5` and low energies `E < 0.7 GeV`, it is not possible to determine the parameter `nu_g` in the bremsstrahlungs parametrization. In that case, the value is set as `nu_g = 0.5`. 
In the propagation tool MUSIC, this is the standard value for all cases.

### DIFF
--- a/src/PROPOSAL/detail/PROPOSAL/scattering/stochastic_deflection/bremsstrahlung/BremsGinneken.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/scattering/stochastic_deflection/bremsstrahlung/BremsGinneken.cxx
@@ -6,6 +6,7 @@
 #include <cmath> 
 #include <stdexcept>
 #include <algorithm>
+#include <string>
 
 using namespace PROPOSAL;
 using namespace std;
@@ -15,11 +16,15 @@ double stochastic_deflection::BremsGinneken::f_nu_g(double nu, double n, double 
 }
 
 double stochastic_deflection::BremsGinneken::get_nu_g(double E, double n, double k_4) const {
-    auto f = [&, n, k_4](double nu) {
+    auto f = [=](double nu) {
         return f_nu_g(nu, n, k_4);
     };
-    auto nu_g_x = Bisection(f, 0.5, 1., 1e-5, 100);
-    return (nu_g_x.first + nu_g_x.second) / 2;
+    if (f(0.5) * f(1.) < 0) { 
+        // check different sign of the definition range, not possible for e_i < 0.7 GeV
+        auto nu_g_x = Bisection(f, 0.5, 1., 1e-5, 100);
+        return (nu_g_x.first + nu_g_x.second) / 2;
+    } 
+    return 0.5;
 }
 
 double stochastic_deflection::BremsGinneken::get_rms_theta(double e_i, double e_f, double mass, double Z) const {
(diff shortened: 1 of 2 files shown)


sudojan:
What is `nu_g`?
This is also printed in a warning message. While a scientist, in general, will understand what `rms_theta` it is not obvious what `nu_g` means.

Also, regarding the energy loss cross-sections, the parameter `$\nu$` often refers to the absolute energy loss and the parameter `v` to the relative energy loss. I would like to be consistent within PROPOSAL.

src/PROPOSAL/detail/PROPOSAL/scattering/stochastic_deflection/bremsstrahlung/BremsGinneken.cxx
```diff
@@ -15,11 +15,10 @@ double stochastic_deflection::BremsGinneken::f_nu_g(double nu, double n, double
 }
 
 double stochastic_deflection::BremsGinneken::get_nu_g(double E, double n, double k_4) const {
-    auto f = [this, &n, &k_4](double nu) {
+    auto f = [&, n, k_4](double nu) {
```
Jean1995, line 18: Since we only want to capture n and k_4 here, can't we also just write
```
auto f = [=](double nu) {
```
or what is the reason that it is written this way?

src/PROPOSAL/detail/PROPOSAL/scattering/stochastic_deflection/bremsstrahlung/BremsGinneken.cxx
```diff
@@ -40,12 +39,16 @@ double stochastic_deflection::BremsGinneken::get_rms_theta(double e_i, double e_
...
+            if (e_i > 0.7) {    // limit of parametrization
+                nu_g = get_nu_g(e_i, n, k_4);
+            };
+
             auto k_5 = k_4 * pow(nu_g, 1 + n) * pow(1 - nu_g, 0.5 - n);
             rms_theta = k_5 * pow(1 - nu, -0.5);
             if (rms_theta < 0.2) {
                 // If no case matches
-                Logging::Get("proposal.deflection")->warn("BremsGinneken deflection: rms_theta = {} > 0.2, nu_g = {}", rms_theta, nu_g);
+                Logging::Get("proposal.deflection")->warn("BremsGinneken deflection: rms_theta = {} < 0.2, nu_g = {}", rms_theta, nu_g);
```
Jean1995, line 51: What exactly does this warning mean? Is it problematic if we run into this case?

pascalgutjahr:
`nu = (E - E') / (E - m)` is the relative energy loss defined by Van Ginneken. I used `nu` as the relative energy loss in all parametrizations of Van Ginneken. 
If we should use `v` instead of `nu`, I can change it for all parametrizations in another pull request. 

Deflections by bremsstrahlung are parametrized as:
Bildschirmfoto 2022-04-21 um 13 32 24 PM
 
The parameter `k_5` is related to `k_4` by continuity. So there is a `nu_g` (g: Grenzwert) in dependence of the energy.

src/PROPOSAL/detail/PROPOSAL/scattering/stochastic_deflection/bremsstrahlung/BremsGinneken.cxx
pascalgutjahr: I have just tested it and it should work. Is there a difference between 
`auto f = [=](double nu) {`
and
`auto f = [&](double nu) {` ? 
Both seem to work.

Jean1995: If you use `&`, all variables are captured by reference.
If you use `=`, all variables are captured by value.
See [here](https://docs.microsoft.com/en-us/cpp/cpp/lambda-expressions-in-cpp?view=msvc-170#capture-clause) for example.

Jean1995: I was just a bit confused about the syntax here, but it shouldn't make a difference in this code since all variabled are onl read inside the lambda expression 😄

src/PROPOSAL/detail/PROPOSAL/scattering/stochastic_deflection/bremsstrahlung/BremsGinneken.cxx
pascalgutjahr: In general, this case should not appear at all. I have just added it as a cross check.

Jean1995: In this case, shouldn't we rather throw an exception here, to avoid that the program continues in an undefined state?

pascalgutjahr: Now there is an exception instead of a warning

Jean1995 (approved):

PR #281, opened 2022-04-20
Add InterpolationSettings::NODES_UTILITY

Jean1995:
We have the nodes options in InterpolationSettings to set the number of nodes for out crosssections interpolation tables.
However, the number of nodes for the utility interpolation tables are currently hardcoded.

This MR introduces the option to adapt this parameter in a similar way, using the `InterpolationSettings::NODES_UTILITY` function. The default value is 500 for all utilities, which means that only the default number of nodes in the ContRand interpolant is changed.

The number of nodes has also been added to the hash so that multiple tables with a different number of nodes can exist at the same time.

### DIFF
--- a/src/PROPOSAL/detail/PROPOSAL/Constants.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/Constants.cxx
@@ -30,6 +30,7 @@ unsigned int InterpolationSettings::NODES_DEDX = 500;
 unsigned int InterpolationSettings::NODES_DE2DX = 200;
 unsigned int InterpolationSettings::NODES_DNDX_E = 100;
 unsigned int InterpolationSettings::NODES_DNDX_V = 100;
+unsigned int InterpolationSettings::NODES_UTILITY = 500;
 
 // precision parameters
 const double PROPOSAL::COMPUTER_PRECISION = 1.e-10;
--- a/src/PROPOSAL/PROPOSAL/Constants.h
+++ b/src/PROPOSAL/PROPOSAL/Constants.h
@@ -56,6 +56,7 @@ struct InterpolationSettings {
     static unsigned int NODES_DE2DX;
     static unsigned int NODES_DNDX_E;
     static unsigned int NODES_DNDX_V;
+    static unsigned int NODES_UTILITY;
 };
 
 // precision parameters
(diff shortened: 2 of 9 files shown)


src/PROPOSAL/detail/PROPOSAL/Constants.cxx
```diff
@@ -30,6 +30,7 @@ unsigned int InterpolationSettings::NODES_DEDX = 500;
 unsigned int InterpolationSettings::NODES_DE2DX = 200;
 unsigned int InterpolationSettings::NODES_DNDX_E = 100;
 unsigned int InterpolationSettings::NODES_DNDX_V = 100;
+unsigned int InterpolationSettings::NODES_UTILITY = 500;
```
sudojan, line 33: I am surprised, that the number of interpolation nodes for the propagation integrals is 500.
In the old PROPOSAL this value was 1000, see e.g. in version 3
https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/v3.0.0/public/PROPOSAL/Constants.h#L62
which was for example used in 
https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/v3.0.0/private/PROPOSAL/ProcessCollection.cxx#L587

When did this change?

sudojan: in version 5.2.0 it was still 1000, see
https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/5.2.0/public/PROPOSAL/methods.h#L108

Jean1995: I am not sure why this number changed. In principal, it would be no problem to go back to 1000 nodes. The size of the propagation utility tables is only around 4 KB for 500 nodes, which is negligible compared to to dNdx tables which have a size of around 320 KB.

Jean1995: Yes, setting the number of nodes of 1000 does not change the runtime, as expected

sudojan: how much does this affect the accuracy?
I remember that 100 were the maximum for cross sections since they are two dimensional.
For Continuous Randomization, 100 was not accurate enough, therefore it was set to 200.
However, for the propagation integrals, a high accuracy was targeted. But maybe 500 nodes are already sufficient..
Can you quantify the decrease in accuracy and the gain in runtime when changing the nodes from 1000 to 500?

sudojan: But the question, to change the number of nodes back to 1000 should be tested in another PR. This one is fine, so far.

PR #282, opened 2022-04-20
Catch calculation that can become `nan` in Highland::CalculateTheta0

Jean1995:
For zero displacement, y will be zero, which will create a nan here.
This is now avoided the same way it is avoided in HighlandIntegral::CalculateTheta0.


sudojan:
What happens for the Moliere Parametrization?
We need a Unit Test, that tests this case for all scattering parametrizations.

Jean1995:
> What happens for the Moliere Parametrization? We need a Unit Test, that tests this case for all scattering parametrizations.

For Molière, there is already [this line](https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/087bc58db5f35f5d43c2723cbecbcf4c5a1ebc92/src/PROPOSAL/detail/PROPOSAL/scattering/multiple_scattering/Moliere.cxx#L62) which checks for a `nan` (which would appear if the grammage equals zero), and returns zero scattering if detected.

I fully agree that I should add a UnitTest for this behaviour.

Jean1995:
I've added a UnitTest that checks for this behaviour. On the master branch, this UnitTest failed with

```
Expected equality of these values:
  offset.sx
    Which is: nan
  0
```
 for Highland scattering

sudojan (approved):

PR #293, opened 2022-06-15
Add secondaries documentation

Jean1995:
While the documentation of our configuration file is quite detailed, there has been only very little documentation of the `Secondaries` class, which is the output of the `Propagator` class. 
However, this class is very important for almost every user.

Therefore, I wrote some documentation for all relevant methods to both C++ and Python.

Furthermore, I added two new functions: `final_state()` and `initial_state()` to get the initial and final particle state of a particle out of the Secondaries class. They are probably more intuitive to use than `track()[-1]` and `track()[0]` which had to be used before.
The jupyter notebook have been adapted accordingly to use these new functions.


Jean1995:
CI is somehow busted at the moment...

sudojan:
> ... get the initial and final particle state of a particle out of the Secondaries class.

You mean "out of the Track"

examples/AdvancedPropagator.ipynb
```diff
@@ -188,7 +188,7 @@
      "name": "stderr",
      "output_type": "stream",
      "text": [
-      "100%|██████████| 100000/100000 [00:17<00:00, 5654.12it/s]\n"
+      "100%|██████████| 100000/100000 [00:01<00:00, 52909.55it/s]\n"
```
sudojan, line 191: is there a reason why it is so much faster?

Jean1995:
> > ... get the initial and final particle state of a particle out of the Secondaries class.
> 
> You mean "out of the Track"

Well, basically both: The Secondaries class contains/manages the track object. However, the idea is that users shouldn't need to understand the track object to get simple information out of the propagator (especially energy losses or the final particle state).

sudojan (approved):

PR #295, opened 2022-06-22
Avoiding unnecessary calculations in `CrossSectionDNDXInterpolant::Calculate`

Jean1995:
### Basics

In PROPOSAL, we have the integral responsible for stochastic loss calculations, which looks like this:

$$\int_{v_\text{min}}^{v} \frac{d\sigma}{dv}dv$$

If we need to calculate the total cross section, we calculate the integral

$$\texttt{dNdx} = \int_{v_\text{min}}^{v_\text{max}} \frac{d\sigma}{dv}dv$$

If we need to sample a stochastic loss, we solve the integral equation

$$\int_{v_\text{min}}^{v} \frac{d\sigma}{dv}dv = \texttt{rnd} \cdot \texttt{dNdx}$$

 for v.

To interpolate the results, we have a two-dimensional interpolation table.
On the first axis of the table, we have the energy $E$.
On the second axis of the table, we have a $v^\prime$, where $v^\prime$ is a transformation of $v$ so that $v \in [v_\text{min}, v_\text{max}]$ and $v^\prime \in [0, 1]$ (which is mostly a logarithmic transformation).
The values saved in the table correspond to the results of the integral 
$$\int_{v_\text{min}}^{v} \frac{d\sigma}{dv}dv.$$

### The runtime issue

In PROPOSAL 6, we stored the results of 

$$\texttt{dNdx} = \int_{v_\text{min}}^{v_\text{max}} \frac{d\sigma}{dv}dv$$

in a separate interpolation table, in addition to the already existing two-dimensional interpolation table.
In PROPOSAL 7, we decided not to create a separate interpolation table as values of $\texttt{dNdx}$ are already stored in the two-dimensional interpolation table.

When profiling the runtime of the $\texttt{dNdx}$ calculation in PROPOSAL 7, I realized that about 50% of the runtime is spent on calculating $v_\text{min}$ and $v_\text{max}$. This is done because in the current implementation, we calculate $v^\prime$ from $v_\text{min}$ and $v_\text{max}$ to know where we have to evaluate the interpolation tables every time. However, we know that by construction, $v^\prime = 1$. And by taking this knowledge into account, we can avoid the explicit calcluation of $v_\text{min}$ and $v_\text{max}$.

##### Direct evaluation of dNdx

At first, I compared 1e7 evaluations of all dNdx values for the default muon cross sections at 10 different energies.

master: 5052 samples ( with 1696 spend in `CrossSectionDNDX::GetIntegrationLimits`)
PR: 3520 samples

#### Simple propagation

I propagate 1e5 muons of energy 1e6MeV through rock, with an e_cut of 500 MeV.
We see no significant difference here because the calculation of the rates is a negligible part of the runtime.

master: 8528 samples
PR: 8594 samples

#### CORSIKA 8 preliminary shower simulations

When simulating a 10 TeV electron shower in CORSIKA 8 (ecut of 2 MeV), we can see a small but significant improvement of the runtime:

master: 17174 samples
PR: 15101 samples


src/PROPOSAL/detail/PROPOSAL/crosssection/CrossSectionDNDX/CrossSectionDNDXInterpolant.cxx
```diff
@@ -62,8 +62,17 @@ size_t CrossSectionDNDXInterpolant::gen_hash(size_t hash) const {
 
 double CrossSectionDNDXInterpolant::Calculate(double energy)
```
sudojan, line 63: maybe a private function getting the energy and retransformed v, which both `Calculate` functions call, might prevent duplicated code

sudojan (approved):

PR #298, opened 2022-07-05
Add onedimensional rate interpolant

Jean1995:
### Synopsis:

**This PR adds a one-dimensional rate interpolant which stores the sum of all dNdx values in a separate interpolation table. This speeds up the call of `Interaction::MeanFreePath` by a factor of 30.**

### Details

Within CORSIKA 8, we figured out that calling the `Interaction::MeanFreePath` function, and as a consequency performing interpolation evaluations, is a limiting time factor (for a shower with magnetic field enabled, this takes about 20% of the total simulation time, and is a majority of the time spent in PROPOSAL).

At the moment, when calling `Interaction::MeanFreePath`, the two-dimensional interpolants for all interaction types and all components are evaluated individually to calculate `dNdx`. However, we are only interested in the sum of all these `dNdx` values.
Therefore, the idea is to instead save these sums in a one-dimensional interpolation table to avoid all these evaluations.

One can see that this improves the runtime when evaluating `Interaction::MeanFreePath` by about a factor of 30! While half of these evaluations come from calling the `CrossSectionDNDX::GetIntegrationLimits` functions, whose calls are avoided with merging #295, this still corresponds to a runtime improvement of a factor of about 15, so still an order of magnitude.

One remaining question is about the implementation:
PROPOSAL doesn't use the `Interaction::MeanFreePath` function. Instead, PROPOSAL either uses the `EnergyIntegral` or needs to calculate the individual dNdx values to sample an interaction. Therefore, PROPOSAL also doesn't have a use for the rate interpolant.
The way I've implemented it at the moment, the rate_interpolant is created every time an InteractionBuilder object is created. This also means that the rate_interpolant is created every time a Propagator is created, although it isn't used at all.

Should we add an option to enable/disable the creation of the rate interpolant? Or should it just be created every time (since the runtime to create the tables is negligible). Or should it be stored in a completely different object?


sudojan:
> Should we add an option to enable/disable the creation of the rate interpolant? 

I think this is the best way.

sudojan:
Do I understand it correctly, that in the Interpolation tables, the inverse of the Mean Free Path Length is stored?
Why is this the case?

Jean1995:
> Do I understand it correctly, that in the Interpolation tables, the inverse of the Mean Free Path Length is stored? Why is this the case?

Yes. The mean free path can become infinite, which can't be stored in the interpolation table. But the inverse (so the sum of all dNdx) just becomes zero in this case, which we can store.

maxnoe:
>  The mean free path can become infinite, which can't be stored in the interpolation table. 

Why is that the case? That seems like a rather strong limitation.

asandrock:
> > The mean free path can become infinite, which can't be stored in the interpolation table.
> 
> Why is that the case? That seems like a rather strong limitation. How to handle it in the interpolation is not easy, but storing it so the actual pole / limit is available should be possible, right?

Because the cross section can become zero on physical grounds, and then the mean free path is infinite in the strict mathematical sense.

maxnoe:
@asandrock I was asking purely about the technical reason why the interpolation tables does not allow to store the special floating point values.

asandrock:
> @asandrock I was asking purely about the technical reason why the interpolation tables does not allow to store the special floating point values.

I can think of three reasons:

1. When the cross section is zero (as is the case in all parametrizations I can think of), dividing by it would lead to NaN AFAIK, unless we rewrite everything to use a signed zero.
2. Additionally, the interpolation through infinite nodes seems highly nontrivial to me. I am not sure whether cubic_interpolation can cope with that.
3. All usecases for the mean free path that I know of actually use the inverse of the mean free path, e.g. the probability distribution for interaction after distance x with mean free path x_mfp is given by P(x) = (1/x_mfp) exp(-x/x_mfp).

sudojan:
Let's move forward in the discussion, here.
I think it is ok, to interpolate the total rate of the interaction probability instead of the mean free path length, but we have to keep in mind, that the interpolation error may be larger in the 'inverse world'.
To keep our promised interpolation precision, we may want to increase the accuracy of our interpolation.

asandrock:
> Let's move forward in the discussion, here. I think it is ok, to interpolate the total rate of the interaction probability instead of the mean free path length, but we have to keep in mind, that the interpolation error may be larger in the 'inverse world'. To keep our promised interpolation precision, we may want to increase the accuracy of our interpolation.

The relative error on the value and its inverse is the same. As far as I can tell, problems will only arise, if the inverse value is near zero, resulting in a large relative error. In our case that would mean a near-infinite mean free path or near-zero interaction probability, i.e. nothing happening at all.

The only application where I can imagine this might be relevant, is the propagation of neutrinos, which is not yet on the master and a separate problem.

Jean1995:
There is another precision problem:

In PROPOSAL 6, we saved the dNdx values in separate tables for each interaction and each component, which means that no (or almost no) kinks were in the functions that we have to interpolate.

However, the interpolation here includes the sum of all dNdx values, which means that we also have a lot of kinks (at the points where a specific component or interaction becomes zero).

With the current number of nodes for the rate_interpolant (500), the interpolation shows artifacts.

For Photons in air with an e_cut at 2 MeV, the result looks like this:

image

For muons in ice with a v_cut of 0.05, the results looks like this:

image

One could probably improve the issue by using more nodes, but at the kinks, this will always be a problem for this kind of interpolation.

sudojan:
The alternative would be to interpolate each interaction.

Jean1995:
> The alternative would be to interpolate each interaction.

You mean for each interaction, or for each combination of interaction and component?
This would probably defeat the original purpose of saving runtime, although I am not entirely sure how much improvement (if any) would still be left

asandrock:
Is not the problem the same for the single interactions/components? If the cross section is zero below some threshold, the cross section will have a kink at this threshold; as far as I remember, none of the cross sections have a smooth behaviour at the threshold, i.e. such that the first derivative is continuous.

asandrock:
The alternative would probably be to save the thresholds and interpolate between them separately, thus obviating the not continuously differentiable kink.

Jean1995:
I played around with the number of nodes for the rate_interpolant tables, as discussed during the last PROPOSAL meeting. Here are my results:

### Runtime

As expected, the evaluation time is independent of the size of the tables. In all cases, it is much faster compared to the old implementation. The following image shows the time it takes to call the `MeanFreePath` method 1e8 times:

image

### File size

As expected, the size of the tables that need to be created scales linearly with the number of nodes.
Even when using 10000 nodes, the table file has a size of 79 KB, which is small compared to the size of our current 2D dNdx tables (313 KB)

image

### Precision

I am comparing three ways of comparing the mean free path length:

- Using the rate interpolant (rate_interpolant)
- Based on interpolated dNdx values for every interaction type and component, i.e. the way it has been calculated before (dNdx_interpol)
- Based on integrated dNdx values, so no interpolation at all (dNdx_integral)

At first, these are the results for **muons** in ice. 
For 500 nodes, the plot looks like this:

image

We see the overshooting of the rate_interpolant (orange line in the top plot). However, we can also see that comparing our general interpolation and integration of dNdx, there are also differences visible (green and blue line in the top plot, yellow line in the bottom plot). 

If we go to 1000 nodes, we still se the overshooting:

image

For 5000 nodes, the overshooting becomes less severe, although still visible:

image

And for 10000 nodes, the overshooting becomes almost negligible:

image

However, this interpretation is somewhat arbitrary. For the plotting, I evaluate at 1000 different energies, however, when I increases this number, the overshooting becomes visible again.

For **electrons** (in air, 2 MeV cut), there are no problems in the interpolant, which means that even for 500 nodes, the results look good:

image

For **photons** (in air, 2 MeV cut), the situation is comparable to the muons.

With 500 nodes, the interpolation overshooting is visible:

image

For 2000 nodes, the overshooting is still slightly visible:

image

And for 5000 nodes, we seem to have reached an optimal value:

image

### Conclusion

We can see that in principle, there is no problem in having a rate_interpolant table with a lot of nodes. For some ridiculously high number of nodes, we would get to a point where the size of the tables would become a memory problem. 

Therefore, I would make the PROPOSAL to include the rate_interpolant, with an independent setting that lets the user set the number of nodes for the rate interpolant. By default, this setting should probably be very high (10000 nodes?) - Rather a bit too high than too low, as every user can decrease this setting if necessary.

sudojan:
Thank you for these extensive studies.
What does a runtime comparison between `rate_interpolant` and `dNdx_inteprol` look like?

Jean1995:
> Thank you for these extensive studies. What does a runtime comparison between `rate_interpolant` and `dNdx_inteprol` look like?

This is shown in my first plot. `time_integral` (which is a misleading label I used in this plot) means that the dNdx_interpol is called and used to calculate the mean free path. So we have a performance gain of a factor of >15 here.

Issue #262, opened 2022-02-16
Testing on latest Windows


sudojan:
As described in PR #261 the latest Windows version of 2022 is causing problems.
Therefore, the CI only tests on Windows 2019.
As soon as this is fixed, we should again test on the latest version.

Jean1995:
#304 has changes this to the latest windows version again.
However, we had to disable the conan cache to make this work. This is something that should be changes in the future (so this issue can stay open for now)

PR #303, opened 2022-08-01
Update spdlog and set windows-latest for CI

Jean1995:
Our CI is currently down. Neither the gcc builds nor the windows builds are finishing.

It looks like is is possible to fix gcc by updating the version of spdlog in conan to the newest version (there was some conflict/problem with fmt). But using the most recent version of a package should always be out goal.

And furthermore, it looks like the `windows-latest` is finally able to build boost (fixes issue #262), and this also fixes the CI.

### DIFF
--- a/.github/workflows/cpp.yml
+++ b/.github/workflows/cpp.yml
@@ -20,7 +20,7 @@ jobs:
             - os : "ubuntu-18.04"
               compiler : "gcc-5"
               compilerpp : "g++-5"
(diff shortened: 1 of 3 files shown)


Jean1995:
I'll rerun all jobs to see whether the fix is still working

Jean1995:
> I'll rerun all jobs to see whether the fix is still working

Nope. Which probably means that there is something broken with the cache.

PR #305, opened 2022-08-03
Log message when tables are created

Jean1995:
This is a first draft for improving the logging for table creation.

Every time before we create a table (either dndx, dedx, de2dx or a utility interpolant), PROPOSAL will check whether a file under this name already exists. If not, PROPOSAL prints a logging message `Tables not existing and need to be created. This may take some time.` If the tables do exist, PROPOSAL prints a logging message `Tables are existing and are read from file.`

However, with a normal logger, this would mean that we would get dozens of identical logging messages.
To avoid this, I am using `spdlog::sinks::dup_filter_sink_st`, which avoids duplicate logging messages that would occur during a specific time interval (at the moment set to 1 minute). This avoids a ton of logging output, but gives still enough logging output to inform the user that things are happening.

For example, when calling `GetStdCrossSections`, but there are no tables, the output looks like this:

```
[2022-08-03 19:59:56.937] [Crosssection] [info] Building KelnerKokoulinPetrukhin cross section for interaction type Brems.
[2022-08-03 19:59:56.937] [TableCreation] [info] Tables not existing and need to be created. This may take some time.
[2022-08-03 19:59:57.235] [Crosssection] [info] Building KelnerKokoulinPetrukhin cross section for interaction type Epair.
[2022-08-03 20:00:27.754] [Crosssection] [info] Building BetheBlochRossi cross section for interaction type Ioniz.
[2022-08-03 20:00:27.937] [Crosssection] [info] Building AbramowiczLevinLevyMaor97 cross section for interaction type Photonuclear.
[2022-08-03 20:01:29.280] [TableCreation] [info] Skipped 16 duplicate messages..
[2022-08-03 20:01:29.280] [TableCreation] [info] Tables not existing and need to be created. This may take some time.
```

Note that there is also an logging information when we start to build a specific crosssection (this has already been implemented before, but has been suppressed because the default loglevel was warn)

When the tables already exist, we only get this output:

```
[2022-08-03 20:04:07.827] [Crosssection] [info] Building KelnerKokoulinPetrukhin cross section for interaction type Brems.
[2022-08-03 20:04:07.827] [TableCreation] [info] Tables are existing and are read from file.
[2022-08-03 20:04:07.828] [Crosssection] [info] Building KelnerKokoulinPetrukhin cross section for interaction type Epair.
[2022-08-03 20:04:07.847] [Crosssection] [info] Building BetheBlochRossi cross section for interaction type Ioniz.
[2022-08-03 20:04:07.851] [Crosssection] [info] Building AbramowiczLevinLevyMaor97 cross section for interaction type Photonuclear.
```

The questions:
- Should the specific information about what crosssection is currently built be printed on loglevel info (like above), or on debug, so that we only see the info *that* tables are currently built?
- Should we log any other events (like when we are writing to memory instead of to files?)
- Should there even be a logging output when the tables are read from a file, or can this also be set to debug?
- Are the messages clear, or should they be changed?
- Should the default loglevel be `info`? It has been `warn` before, but this would suppress these log messages (unless we change their loglevel to `warn`, but this is not intuitive to me)

### DIFF
--- a/src/PROPOSAL/PROPOSAL/Logging.h
+++ b/src/PROPOSAL/PROPOSAL/Logging.h
@@ -30,6 +30,7 @@
 #include <PROPOSAL/methods.h>
 #include <spdlog/sinks/sink.h>
 #include <spdlog/spdlog.h>
+#include "spdlog/sinks/dup_filter_sink.h"
 
 #include <memory>
 #include <string>
@@ -48,8 +49,12 @@ struct Logging {
     static logger_ptr Get(std::string const& name)
     {
         auto it = Logging::logger.find(name);
-        if (it == logger.end())
-            Logging::logger[name] = Logging::Create(name);
+        if (it == logger.end()) {
+            if (name == "TableCreation")
+                Logging::logger[name] = Logging::CreateDupFiler(name);
+            else
+                Logging::logger[name] = Logging::Create(name);
+        }
         return Logging::logger[name];
     }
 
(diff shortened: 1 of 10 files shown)


Jean1995:
Aparently, the last conan release broke some stuff in the boost conan package (see for example [here](https://github.com/conan-io/conan-center-index/issues/12159)).

Conan people are aware of it so this should be fixed soon.

Jean1995:
This will be merged together with #306

Issue #292, opened 2022-06-15
Interpolation tables are stored as txt files but they are actually binary files


Jean1995:
Forwarded from [this issue](https://gitlab.iap.kit.edu/AirShowerPhysics/corsika/-/issues/500) in CORSIKA 8.

We save our interpolation tables as .txt files, which implicated that they include readable text. However, they actually include binary data.

Jean1995:
Fixed by merging #306

Issue #290, opened 2022-05-09
Logging information if tables are being built


Jean1995:
We may want to add some information for the user if PROPOSAL is currently building tables.

We already have some detailed information about tables being built when one uses `debug` logging mode, but especially for new users, it might be useful to have some information like:

"PROPOSAL tables are being built for the first time. This might take some time..."

Jean1995:
Fixed by merging #306

PR #306, opened 2022-08-11
Logging for table creation and change file extension of tables from .txt to .dat

Jean1995:
Fixes issues #292 and and #290.

- Use .dat instead of .txt to show that we are actually storing binary files, not readable files.
- Log when tables are not found and need to be created (log level warn), only do this once 
- Log when tables are not found and need to be stored in memory (log level warn), only do this once


src/PROPOSAL/PROPOSAL/methods.h
```diff
@@ -34,8 +34,11 @@
 #include <memory>
 #include <vector>
 #include <sys/stat.h>
+#ifdef _WIN32
```
sudojan, line 37: Why is this necessary?
a comment would be helpful

sudojan:
setting the default log level from warning to info should be mentioned in the PR description

src/PROPOSAL/PROPOSAL/methods.h
Jean1995: unistd.h is a header that provides access to the API of the operating system, in this case to check paths (this header has also been used in older PROPOSAL versions).
However, unistd.h is by default not available for Windows. But for Visual Studio, io.h can be used as a replacement for unistd.h.

I am not 100% sure whether this fix works for all Windows systems, or only for those using Visual Studio. I should probably have a closer look at this. Nevertheless, I will add a comment to this for clarification.

Jean1995: Note that this is actually not part of this PR, but part of the PR #305. I've just rebased this branch onto the branch of PR #305, so the changes are shown here.

Jean1995:
> setting the default log level from warning to info should be mentioned in the PR description. Are there further changes?

This is also part of PR #305. Therefore I will also answer your question about the minor changes in the corresponding PR.

maxnoe:
The default logging level should be warning! That's a very sensical default. By default, only things going wrong / requiring user interaction should warn.

Jean1995:
> The default logging level should be warning! That's a very sensical default. By default, only things going wrong / requiring user interaction should warn.

The background behind this has been that users (especially CORSIKA 8) have been complaining that when PROPOSAL starts to build tables (which can take some minutes), there is no indication that things are happening (and people thought the program might be stuck). 
Therefore, we added a  logging information about tables being built. But this is, in my opinion, nothing that should be logged as log_warn, but rather as log_info. Therefore, we changed the default to log_info so that people are actually (by default) informed about the fact that tables are currently built.

I am not sure how to combine this request with having warn as a default loglevel.

maxnoe:
Something taking a long time might warrant "warning". I'd also rather have a log message about which specific table is being built. That would also not require "deduplicating" warning messages.

Jean1995:
> Something taking a long time might warrant "warning". I'd also rather have a log message about which specific table is being built. That would also not require "deduplicating" warning messages.

The log message about the specific tables being built are, at the moment, logged when loglevel debug is used. 
Per default, I think that this information shouldn't be printed every time (so it should be below the default loglevel). Otherwise, this can be quite some spam.

This is why I have introduced this separate, general message about tables being built. We can use log warn for this if we want to keep the default log level as warn. Is warn always the default setting for software projects?

> For CORSIKA users, it might also be useful to log something like:
> 
> `"Running PROPOSAL for the first time, creating interpolation tables. To avoid this, copy prebuilt interpolation tables into <DIRECTORY>`".

Yes, something like this would be optimal. This is what I wanted to achieve. But in general, it is nor clear when this message should appear. This is why I used the approach with the specific logger that avoids duplicate messages.

> CORSIKA might also set the logging level of that specific logger to INFO.

Yes, this is of course an option. However, I don't think that the behavior of standalone PROPOSAL and PROPOSAL within CORSIKA needs to be different.

Jean1995:
I've made some adaptions from the comments.
Warn is now the default loglevel again. Information about missing tables are now logged as log::warn. 
Instead of using the specific logger that avoids duplicate messages, we will now check whether we have already issued a warning for a specific tables path.

sudojan (approved):

Issue #301, opened 2022-07-22
Change LowerEnergyLimit of ComptonEffect


Jean1995:
As the moment, we have set the lower limit for the Compton effect crosssection to 2*ME. This has been a rather arbitrary cut we chose since this is the lowest energy where electron-positron pairs can still be produced.

However, as soon as we merge the photoeffect (#296), we can go to lower energies than 2*ME. The question is now which now lower energy limit should be introduced.

Jean1995:
One idea from @asandrock:

We might use the ionization energy of the K-shell electron. This is the LowerEnergyLim used by the new implementation of the photoeffect.  Since we can't describe photons below this energy anyway (at least for now), this would be a possible choice for a new, "arbitrary" lower limit.

Jean1995:
Fixed with PR #307

PR #307, opened 2022-08-11
Adapt LowerEnergyLim of Compton

Jean1995:
Fixes issue #301 

The lower energy limit of Compton is somewhat arbitrary, and has been set to 2*ME before. Now, we set it to the ionization energy of the K-shell electron of hydrogen to have a new lower limit which works better with the now-implemented Photoeffect.

### DIFF
--- a/src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/Compton.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/Compton.cxx
@@ -11,7 +11,9 @@ using namespace PROPOSAL;
 double crosssection::Compton::GetLowerEnergyLim(
     const ParticleDef&) const noexcept
 {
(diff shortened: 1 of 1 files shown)


src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/Compton.cxx
```diff
@@ -11,7 +11,7 @@ using namespace PROPOSAL;
 double crosssection::Compton::GetLowerEnergyLim(
     const ParticleDef&) const noexcept
 {
-    return 2. * ME;
+    return pow(ALPHA, 2) * ME / 2; // ionization energy of K-shell electron for hydrogen
```
sudojan, line 14: compared to the lower limit of the photo effect parametrization of $(Z\alpha)^2 m_e/2$ this now ensures that the Compton lower limit is always at least as low as the photo effect.

sudojan (approved):

PR #308, opened 2022-08-16
Change kinematic limits of BremsElectronScreening

Jean1995:
For BremsElectronScreening, the kinematic limits have so far been taken from Petrukhin & Shestakov, where the upper limit has been calculated by setting the corresponding screening function to zero.

However, we don't need to use this for BremsElectronScreening. Instead, we can just use the physical limit 1 - m/E. 
We have checked that this is still consistent with the condition that the screening function should be non-negative (see [this presentation](https://indico.scc.kit.edu/event/2955/contributions/11154/attachments/5345/8203/EM_comparison.pdf) for more information)

Comparing PROPOSAL with EGS4, using these improved limits also yields a better agreement in the cross section (also shown in the presentation linked above).

This PR also fixes that the tables with the correction factors are stored as kinetic energies, but have been evaluated with total energies so far.

### DIFF
--- a/src/PROPOSAL/PROPOSAL/crosssection/parametrization/Bremsstrahlung.h
+++ b/src/PROPOSAL/PROPOSAL/crosssection/parametrization/Bremsstrahlung.h
@@ -77,7 +77,7 @@ namespace crosssection {
 
         double GetLowerEnergyLim(const ParticleDef&) const noexcept final;
         KinematicLimits GetKinematicLimits(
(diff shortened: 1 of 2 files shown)


sudojan:
Do I understand it correctly, that there are default limits for interactions, e.g. for Bremsstrahlung they are 0 and $1- m/E$, which can be overridden if a parametrization defines its own limits, which always have to be inside the default limits?

asandrock (approved):

PR #312, opened 2022-08-18
Add missing entry for InteractionType::Photoeffect

Jean1995:
Add missing entry for InteractionType::Photoeffect to Type_Interaction_Name_Map

Without this, it is currently impossible to search for Photoeffect losses in `Secondaries::GetStochasticLosses(const std::string& interaction_type)`,  so that one has to use `Secondaries::GetStochasticLosses(const InteractionType& type)` instead.

Most of the other uses of this map affect logging messages for building tables - But there are no tables built for the photoeffect.


Jean1995:
Apparently, conan-center-index (again) broke something with the boost recipe 😅 ([this PR](https://github.com/conan-io/conan-center-index/pull/11988#issuecomment-1218367469))

I hope that this will be fixed soon

sudojan (approved):

PR #314, opened 2022-08-19
Made correction factor non-relativistic electrons numerically stable at very low energies

asandrock:
Using the approximation sinh(x) ~ exp(x)/2 for large values of x and combining this with the argument of the exponential function already present, the floating point exception from overflow is averted. This approximation is accurate to 1e-9 for x > 10, which is much more accurate than the photoeffect cross section parametrization itself.


Jean1995:
LGTM. I have also used this in CORSIKA 8, there are no more errors with this fix.

Jean1995 (approved):

PR #315, opened 2022-08-19
Update ubuntu-18.04 runner to ubuntu-20.04

Jean1995:
The ubuntu-18.04 runner of GitHub actions starts deprecation in a few days (which will include temporary brownouts), with complete deprecation at the end of this year (see [this issue](https://github.com/actions/runner-images/issues/6002)).
Therefore, we should update our runners.

Since gcc5 is not trivially available on 20.04 anymore, it is now downloaded and installed manually. We should still test gcc5 because this is the oldest gcc version that is still tested by conan.

For gcc7, we can use a simpler tool that installs and initializes gcc7.

### DIFF
--- a/.github/workflows/cpp.yml
+++ b/.github/workflows/cpp.yml
@@ -26,6 +26,11 @@ jobs:
         CXX: ${{ matrix.compilerpp }}
       steps:
         - uses: actions/checkout@v2
+        - name: add custom deb repository
+          if: ${{ matrix.compiler == 'gcc-5' }}
+          uses: myci-actions/add-deb-repo@10
+          with:
+            repo: deb http://dk.archive.ubuntu.com/ubuntu/ xenial main
         - name: Install gcc-5
           if : ${{ matrix.compiler == 'gcc-5' }}
           run: |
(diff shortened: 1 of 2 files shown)


sudojan:
Since Ubuntu 20 ships with python 3.8, should we also increase the python version from 3.7?

sudojan:
> We should still test gcc5 because this is the oldest gcc version that is still tested by conan.

Do we really need to do this? At least when we want to switch to cpp17 (maybe in PROPOSAL8) this will become necessary to drop.

Jean1995:
> > We should still test gcc5 because this is the oldest gcc version that is still tested by conan.
> 
> Do we really need to do this? At least when we want to switch to cpp17 (maybe in PROPOSAL8) this will become necessary to drop.

At the moment, conan is testing to PROPOSAL on gcc5. However, we might drop the support for gcc5 in the conan-center-index receipe. This wouldn't be a problem for CORSIKA 8, since they are cpp17 anyway and therefore don't support gcc5.

sudojan (approved):

PR #317, opened 2022-09-19
Use closed form for AnnihilationHeitler parametrization

Jean1995:
### Motivation 

In the current version 7.4.1 of PROPOSAL, we still used the old `2*ME` limit for `GetLowerEnergyLim`:

https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/6791de1bb3eac2a7b3a3e992ba79e35e1d44c839/src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/Annihilation.cxx#L15-L19

This means that for the AnnihilationInterpolant (solid line in the plot), the cross sections abruptly becomes zero for `E = 2*ME`, while for AnnihilationIntegral (dashed line in the plot), the cross section is continued to `E=ME`:

anni1

This also means that the rate_interpolant (i.e. the sum of all dNdx) is wrong, as it becomes zero too early:

anni2
This also caused the rate_interpolant to become negative, which caused errors in CORSIKA 8 (as there has been no check for negative rate_interpolant resuls).

However, just changing the  `GetLowerEnergyLim` to ME is not a valid solution, as the Annihilation cross section diverges for E->m as 1/sqrt(E - ME). If one tries to interpolate the cross section this way, the result looks like this (the dashed red lines indicate E=ME and E=2*ME):

image

However, since we know a closed form of the Heitler cross section, we don't actually need to use interpolation tables for dNdx. Since we have implemented `ParametrizationDirect` as a class with 7.2.0, we can use the same design that we have already used for the Photoproduction and the Photoeffect also for Annihilation. This is done in this PR.

### Implementation and Validation

I used pen and paper to integrate the differential crosssection to obtain a closed form of the total crosssection (this result has been verified with the EGS and Geant4 manual). The UnitTests confirm that we still get the same results, apart from numerical uncertainties at very-high energies (E > 1e10 MeV).

Concerning the performance, I've profiled the evaluation of `CalculatedNdx`, and the evaluation of the new, closed form is simple enough to even get a small performance gain compared to old evaluation where tables have been used (~`6000` samples on the master, ~`4145` samples with this PR)

While the divergence for E->ME is still there, it still needs to be interpolated for the rate_interpolant.
However, we use enough interpolation nodes (by default 10000) to be able to interpolate around the divergence. Only directly at the divergence, as well as around the regime where ionization becomes zero, the rate_interpolant is slightly off:

anni2

anni3

Another point that is changing is the calculation of the secondary particles, where the asymmetry parameter between the created photons (`rho`) needs to be sampled.
In the old PROPOSAL implementation, this has been done by sampling from the differential cross section. Now that there is no class providing the differential cross sectionanymore, this sampling is directly done within the secondaries class. Instead of integrating the differential cross section, a closed form of the equation $\int_{\rho_{min}}^{\rho} \frac{d\sigma}{d\rho} = \xi_{rnd} \int_{\rho_{min}}^{\rho_{max}} \frac{d\sigma}{d\rho}$ is solved for $\rho$, which is done by using the bisection method.

Regarding the performance, this is even faster than the previous method (~3400 samples vs. ~6330 samples).
Comparing the old and new version result-wise, they agree quite well:

anni4

I've also added a few extra UnitTest to validate the results.


sudojan:
> The UnitTests confirm that we still get the same results, apart from numerical uncertainties at very-high energies (E > 1e10 MeV).

Can you plot the uncertainties over the complete energy range, 1-1e14 MeV?

Jean1995:
> > The UnitTests confirm that we still get the same results, apart from numerical uncertainties at very-high energies (E > 1e10 MeV).
> 
> Can you plot the uncertainties over the complete energy range, 1-1e14 MeV?

anni5

This compares the new implementation (dNdx_new) with the old implementation (dNdx_interpol and dNdx_integral).

sudojan:
First, this looks acceptable.
But, I wonder why there are oscillations also at the integral?

Jean1995:
> First, this looks acceptable. But, I wonder why there are oscillations also at the integral?

This is because integrating over `rho` numerically, which id done for dNdx_integral, becomes numerically unstable. 
To verify that these oscillations are from the old method, and not the new method, this is a zoom into the dNdx values at very-high energies:

anni6

sudojan (approved):

PR #319, opened 2022-09-21
Check if dE2dx tables have been built when creating ContRand builder

Jean1995:
When creating a EnergyCutSettings object, a boolean `continuous_randomization` indicated whether dE2dx tables should be built.
However, one can still initialize a ContRand object if this option has been set to false. In this case, the object will exist, but actually make no continuous randomization. 
This is not useful, and is probably not wanted by the user. Therfore, this PR adds a check whether at least one dE2dx table has been built, and if this is not the case, an exception is thrown:

```
Can not build ContRand because no dE2dx tables have been built. You need to use continuous_randomization=True when creating the EnergyCutSettings object to be able to use continuous randomization.
```

This also means that the ContRand object needs to know about the EnergyCutSettings that are being used by the CrossSection objects. Therefore, a `GetEnergyCutSettings` method has been added to the CrossSection classes.

### DIFF
--- a/tests/ContinuousRandomization_TEST.cxx
+++ b/tests/ContinuousRandomization_TEST.cxx
@@ -27,11 +27,22 @@ ParticleDef getParticleDef(const std::string& name)
 
 TEST(ContinuousRandomization, Constructor)
 {
-    auto p_def = MuMinusDef();
+    auto p_def = EPlusDef();
     auto medium = Ice();
     auto cuts = std::make_shared<EnergyCutSettings>(INF, 1, true);
     auto cross = GetStdCrossSections(p_def, medium, cuts, false);
     auto cont_rand = make_contrand(cross, false);
+}
+
+TEST(ContinuousRandomization, ConstructorNoTables)
+{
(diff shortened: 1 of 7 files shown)


tests/ContinuousRandomization_TEST.cxx
```diff
@@ -27,13 +27,24 @@ ParticleDef getParticleDef(const std::string& name)
 
 TEST(ContinuousRandomization, Constructor)
 {
-    auto p_def = MuMinusDef();
+    auto p_def = EPlusDef();
```
sudojan, line 30: Was there a reason to change the particle?

Jean1995: Because EPlusDef, per default, includes an only-stochastic cross section (annihilation), while muons do not.

This is basically a check for [this line](https://github.com/tudo-astroparticlephysics/PROPOSAL/pull/319/files#diff-751770671f6e83fb8810b1b206768de58df5361aa839a3120e2b8888bd256a5aR18) in the code.

sudojan (approved):

PR #325, opened 2022-11-08
Allow installing proposal via `conan build`

maxnoe:
See documentation here for intended use:

https://docs.conan.io/en/latest/reference/conanfile/attributes.html?highlight=should_install#should-configure-should-build-should-install-should-test


Jean1995 (approved):
Sorry for the late review, but this looks fine to me

PR #328, opened 2022-11-15
Fix clang warning due to missing override keyword

Jean1995:
When building PROPOSAL, clang currently gives a bunch of warnings such as 

```
warning: 'CalculatedNdx' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
        double CalculatedNdx(double energy, size_t hash) {
```

which is due to missing `override` keyword in CrossSectionDirect.h.
This is for example visible in [this runner](https://github.com/tudo-astroparticlephysics/PROPOSAL/actions/runs/3418261048/jobs/5690338119).

This commit should fix them.


Jean1995:
Warnings in the clang build are now gone, so this should be fine

sudojan (approved):

PR #334, opened 2023-01-04
Add LPM effect for photopair, fixes for BremsLPM

Jean1995:
This PR adds the LPM effect for pair production of electron-positron pairs by photons.

Furthermore, this PR fixes the following issues in the implementation of the LPM effect in Bremsstrahlung:

1. An inconsistent definition of eLPM has been fixed
2. A more general definition of s1 is now used, to that the BremsLPM parametrization is not only valid for muons, but also for other particles (electrons/positrons/taus)

### DIFF
--- a/docs/config_docu.md
+++ b/docs/config_docu.md
@@ -446,6 +446,7 @@ For there parametrizations, the parametrization of the shadowing effect can be c
 | ----------------- | ------- | ------- | -------------------------- |
 | `parametrization` | String  | `-`     | Parametrization to be used |
 | `multiplier`      | Number  | `1.0`   | Multiplier to scale the cross section with a coefficient |
+| `lpm`             | Boolean | `true`  | Enabling or disabling the reduction of the cross section at high energies by the Landau-Pomeranchuk-Migdal effect.  |
 
 Creation of an electron-positron pair by an ingoing photon. Available `photopair` parametrizations are:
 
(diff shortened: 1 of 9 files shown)


Jean1995:
Looks like something odd is happening to the authorship of the merge commits in this PR.

maxnoe:
@Jean1995 Why? It looks like @asandrock just rebased and thus he commited (again, via rebase) your commits, so these have now commiter info @asandrock and author info @Jean1995

maxnoe:
But these are very many commits for just a relatively small change, this should probably be rebased against current master again and/or squashed into a single / few commits.

Jean1995:
> @Jean1995 Why? It looks like @asandrock just rebased and thus he commited (again, via rebase) your commits, so these have now commiter info @asandrock and author info @Jean1995

True, I didn't realize that only the committer name changed.

Jean1995:
> But these are very many commits for just a relatively small change, this should probably be rebased against current master again and/or squashed into a single / few commits.

There are definitely several changes that should be kept separate into different commits. If it is possible to squash commits into more than just one commit, this might make the commit history more straight-forward.

maxnoe:
> There are definitely several changes that should be kept separate into different commits. If it is possible to squash commits into more than just one commit, this might make the commit history more straight-forward.

One easy way to achieve this is to do a `git fetch`, `git reset --soft origin/master`, `git reset` and then use `git add -p` / `git commit` to create several selective commits.

Jean1995:
> .... and you can start making nice small commits.

Won't this process remove the original author information? Or can I just change that the the commit author using `git commit --amend --author"`?

maxnoe:
> Won't this process remove the original author information? 

It will remove any previous commit, converting all commits to uncommited changes locally. So yes.

> Or can I just change that the the commit author using git commit --amend --author"?

You can always the commiter id for a commit by using `git -c user.name="xxx" -c user.email="..." commit` or you can add a 
`Co-authord-by: Name <email>` line at the bottom of the commit message for multiple authors.

Jean1995:
Closed in favor of #336 where the commit history has been cleaned up.

PR #336, opened 2023-01-04
Add LPM effect for photopair, fixes for BremsLPM 

Jean1995:
New PR for #334 but with clean history.

This PR adds the LPM effect for pair production of electron-positron pairs by photons.

Furthermore, this PR fixes the following issues in the implementation of the LPM effect in Bremsstrahlung:

1. An inconsistent definition of eLPM has been fixed
2. A more general definition of s1 is now used, to that the BremsLPM parametrization is not only valid for muons, but also for other particles (electrons/positrons/taus)

### DIFF
--- a/src/PROPOSAL/PROPOSAL/crosssection/parametrization/PhotoPairProduction.h
+++ b/src/PROPOSAL/PROPOSAL/crosssection/parametrization/PhotoPairProduction.h
@@ -34,9 +34,14 @@
 
 namespace PROPOSAL {
 namespace crosssection {
+    class PhotoPairLPM;
+
     class PhotoPairProduction : public Parametrization<Component> {
+    protected:
+        std::shared_ptr<PhotoPairLPM> lpm_;
+        double density_correction_; // correction to standard medium density for LPM
     public:
-        PhotoPairProduction() = default;
+        PhotoPairProduction();
         virtual ~PhotoPairProduction() = default;
 
         double GetLowerEnergyLim(const ParticleDef&) const noexcept final;
@@ -49,7 +54,9 @@ namespace crosssection {
     };
 
     struct PhotoPairTsai : public PhotoPairProduction {
-        PhotoPairTsai() { hash_combine(hash, std::string("tsai")); }
+        PhotoPairTsai(bool lpm = false);
+        PhotoPairTsai(bool lpm, const ParticleDef&, const Medium&,
+                          double density_correction = 1.0);
         using base_param_t = PhotoPairProduction;
         std::unique_ptr<Parametrization<Component>> clone() const final;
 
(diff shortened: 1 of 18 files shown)


Jean1995:
Enclosed are some exemplary comparisons of Bremsstrahlung dEdx between this PR and the current master:

brems_MuMinus_ice

brems_EMinus_lead

brems_TauMinus_ice

tests/gen_testfiles_scripts/photopair.py
```diff
@@ -35,32 +37,34 @@ def create_tables(dir_name, **kwargs):
...
-                        buf[key][1].append(particle.name)
-                        buf[key][1].append(medium.name)
-                        buf[key][1].append(str(multiplier))
-                        buf[key][1].append(str(energy))
-                        buf[key][1].append(xsection.param_name)
-                        buf[key][1].extend(result)
-                        buf[key][1].append("\n")
-
-                    buf[key][0].write("\t".join(buf[key][1]))
+                for lpm in lpms:
```
sudojan, line 40: maybe an additional loop with densities 1 and 2 might be useful to test the LPM calculation?

src/PROPOSAL/PROPOSAL/crosssection/parametrization/PhotoPairProduction.h
```diff
@@ -86,5 +95,21 @@ namespace crosssection {
     template <> struct ParametrizationId<PhotoPairKochMotz> {
         static constexpr size_t value = 1000000013;
     };
+
+    // LPM effect object
+    class PhotoPairLPM {
```
sudojan, line 100: do we really need this constructor with `mol_density`, `mass_density`, `sum_charge`, `eLPM`? Some variables are not really intuitive.

Jean1995: `eLPM` is a computationally expensive calculation that needs to be executed only once, so we should save it.
`mol_density_` and `sum_charge_` are medium variables. Since those are the only two variables we need for the calcluations, we save them to avoid saving the whole `medium` in the PhotoPairLPM object. So I believe these three values are useful to save. Their names are just adopted from what they are called in the `Medium` class (although I agree that the variable names are not always well-chosen, but for the sake of consistency, I would use the same names here).

`mass_density_` is only used in the constructor, so I would agree that we don't need to save them as a private member of the object. But in my opinion, this is the only change I would make to the constructor.

Jean1995:
As an additional validiation, I have created plots where I compare the new dEdx results to the results from [Polityko et al](https://iopscience.iop.org/article/10.1088/0954-3899/28/3/306):

brems_MuMinus_iron

brems_MuMinus_water

Note that the energies used in their paper only go up to 1e18 MeV, but it is sufficient to see that the agreement is now much better.

sudojan (approved):

PR #345, opened 2023-02-08
Remove extra sampling of asymmetry in photopairproduction secondary energy calculation

Jean1995:
I noticed that in `PhotoPairProductionInterpolant::CalculateEnergy`, there is a step where we "randomize" how we distribute the energy between the electron and positron.

However, for this parametrization, `rho` (or in literature rather known as `x`, the ratio of the electron and photon energy) is already between [0, 1). This means we don't need to randomize the distribution.

This is probably a remnant from other classes, where `rho` is just sampled between [0, 1) due to the symmetry of the function. However, this is not implemented for PhotoPairProduction (although it could be implemented).

Therefore, we don't need 5, but 4 random numbers here.

### DIFF
--- a/src/PROPOSAL/PROPOSAL/secondaries/parametrization/photopairproduction/PhotoPairProduction.h
+++ b/src/PROPOSAL/PROPOSAL/secondaries/parametrization/photopairproduction/PhotoPairProduction.h
@@ -18,8 +18,7 @@ namespace secondaries {
         virtual std::tuple<Cartesian3D, Cartesian3D> CalculateDirections(
             const Vector3D&, double, double, const Component&, double, double,
             double) = 0;
-        virtual std::tuple<double, double> CalculateEnergy(double, double, double)
-            = 0;
+        virtual std::tuple<double, double> CalculateEnergy(double, double) = 0;
     };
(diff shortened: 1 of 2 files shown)


sudojan:
So the questions are,
- is the distribution symmetric in rho? 
- if the above is the case, is the distribution, we need to sample from, also much easier when just considering the range between 0 and 0.5? (which is e.g. the case for the PairProduction induced by muons, where it greatly improves the accuracy).
- is the increase in accuracy worth it to draw an additional random number?

But for now, you are right, it does not make sense to draw this additional random number.

Jean1995:
> * is the distribution symmetric in rho?

While the KochMotz parametrization is completely symmetric in rho, the parametrization of Tsai surprisingly isn't. It is for E >> m_e, but otherwise, there is an asymmetry. It appears when transforming the differential cross section from dsigma/dp (with p the absolute electron momentum) to dsigma/dx. 

I have to think a bit more about why this is happening.

I would open a new issue about the idea of sampling rho between 0 and 0.5.

sudojan (approved):
Since it is shown, that there can be an asymmetry, this can be merged

PR #346, opened 2023-02-08
Add new sampling method for photopair

Jean1995:
When investigating the two photopair angle sampling methods we currently use (Tsai and ForwardPeaked), I saw that the median value differs by quite a lot:

image

I looked in the EGS5 manual, where a very simple sampling method has been suggested with formula (2.157).
I implemented this method as a test, and to my surprise, it reproduces the current implementation by Tsai quite well:

image

Note that both plots are created for an interaction of a 1e4 MeV Photon with a nitrogen atom.

However, the Sauter implementation is almost as quick as the crude EGS4 approximation, which itself is two order of magnitudes quicker than the Tsai sampling.

I am a bit surprised that the Sauter sampling methods reproduces the results by Tsai that well, especially since in the EGS5 manual, the method is called "a crude approximation even in its region of validity". 
This good agreement also does not change for different Z or energies.
So probably, the Tsai calculation is much worse than I thought, or the Sauter approximation much better than they say in the EGS manual.

So maybe someone should investigate this a bit more before merging?

### DIFF
--- a/src/PROPOSAL/PROPOSAL/secondaries/parametrization/photopairproduction/PhotoPairProductionKochMotz.h
+++ b/src/PROPOSAL/PROPOSAL/secondaries/parametrization/photopairproduction/PhotoPairProductionKochMotz.h
@@ -2,6 +2,7 @@
 
 #include "PROPOSAL/secondaries/parametrization/photopairproduction/PhotoPairProductionInterpolant.h"
 #include "PROPOSAL/crosssection/parametrization/PhotoPairProduction.h"
+#include "PROPOSAL/secondaries/parametrization/photopairproduction/SauterSampling.h"
 
 namespace PROPOSAL {
     namespace secondaries {
@@ -16,7 +17,7 @@ namespace PROPOSAL {
 
         };
 
-        class PhotoPairProductionKochMotzSauter : public PhotoPairProductionKochMotzForwardPeaked {
+        class PhotoPairProductionKochMotzSauter : public PhotoPairProductionKochMotzForwardPeaked, public SauterSampling {
         public:
             PhotoPairProductionKochMotzSauter() = default;
             PhotoPairProductionKochMotzSauter(ParticleDef p, Medium m)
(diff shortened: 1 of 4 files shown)


src/PROPOSAL/PROPOSAL/secondaries/parametrization/photopairproduction/PhotoPairProductionTsai.h
```diff
@@ -30,5 +30,43 @@ namespace secondaries {
                 const Vector3D&, double, double, const Component&,
                 double, double, double) override;
     };
+
+    class PhotoPairProductionTsaiSauter : public PhotoPairProductionTsaiForwardPeaked {
```
sudojan, line 34: Is this the exact same as `PhotoPairProductionKochMotzSauter` and if yes, can it be implemented to avoid duplication?

src/PROPOSAL/PROPOSAL/secondaries/parametrization/photopairproduction/PhotoPairProductionKochMotz.h
```diff
@@ -15,5 +15,43 @@ namespace PROPOSAL {
...
+
+                // Sampling according to EGS 5 manual, formula (2.157)
+
+                auto energies = CalculateEnergy(energy, rho);
+                double E_electron = std::get<0>(energies);
+                double E_positron = std::get<1>(energies);
+                double p_electron = std::sqrt(E_electron * E_electron - ME * ME);
+                double p_positron = std::sqrt(E_positron * E_positron - ME * ME);
+
+                // TODO: should this be two times the same random numbers, or two different ones?
```
sudojan, line 38: Have you tried both and did you see a bias when using the same rnd?
I would personally say, that this shouldn't be independent.

Jean1995: I've had a look at the source code of EGS5, and there, they use two independent random numbers for the sampling procedure.

(For the record: I found this in `egs5_pair.f`, lines 95 and following.)

sudojan: ok, then let|s make it consistent with EGS and leave it like this

sudojan (approved):

Jean1995:
One remaining question I have: Currently, we still use `m/E`  (from EGS4) as the default sampling method.

Do we want to change the default to `Sauter` (because it produces better results, with a similar computational effort), or we want keep to keep the EGS4 approximation as a default?

sudojan:
> One remaining question I have: Currently, we still use `m/E` (from EGS4) as the default sampling method.
> 
> Do we want to change the default to `Sauter` (because it produces better results, with a similar computational effort), or we want keep to keep the EGS4 approximation as a default?

Good point, that I forgot.
yes we should move to the default to Sauter

sudojan (approved):

PR #353, opened 2023-02-17
Add new parametrizations of photoproduction based on CORSIKA 7

Jean1995:
This PR adds two parametrizations of photoproduction based on the implementation used in CORSIKA 7.
The parametrization is based on the implementation of the continuum of the photon-proton cross section from [here](https://web.iap.kit.edu/heck/publications/kit-swp-6_vectormeson.pdf). Note that this is different from the parametrization given in the CORSIKA 7 physics reference manual, which appears to be outdated.

Since we were unable to implement the parametrization of the resonances that are listed, instead, we use the implementation of the three highest resonances in [SOPHIA](https://arxiv.org/pdf/astro-ph/9903478.pdf). 

One can compare the implementation using the SOPHIA resonances with Figure 2 from [here](https://web.iap.kit.edu/heck/publications/kit-swp-6_vectormeson.pdf), and can see that the agreement is good enough

image.

To describe the shadowing effect, the parametrization given [here](https://web.iap.kit.edu/heck/publications/kit-swp-6_vectormeson.pdf) uses the approximation `sigma_{\gamma, A} = sigma_{\gamma, p} * A**0.91`. This has been implemented at `HeckC7Shadowing`.

As an alternative, the `sigma_{\gamma, p}` cross section with the shadowing by Bezrukov and Bugaev (which is also used in PROPOSAL for the other photoproduction parametrizations) is implemented as `Heck`.

Note that this causes a significant difference in the cross section, both around the resonances and in the high-energy region. This plots shows the two parametrizations, together with the `Rhode` parametrization:

image

One can see that the Heck parametrization and the Rhode parametrization have a good agreement at the resonances, but only if the same shadowing parametrization is used.

### What else?

- In the future, it might be good to implement the possibility to use different shadowing parametrizations for the photoproduction parametrizations, the same way it is already possible for the photonuclear parametrizations. But this would require some additional restructuring out of the scope of this PR.
- We might think about which parametrization we want to use as a default. Currently, this is Rhode. One could switch to `Heck` as a default if we want to have the different high-energy extrapolation.
- I am not 100% convinced that the names of the parametrizations are well-chosen.


asandrock:
For comparison to C7, I think it is a good idea to keep this parametrization with HeckC7Shadowing.

The extrapolation to high energies is an open research problem. There are arguments that a Pomeron-like behaviour $\sigma \propto s^{1 + \epsilon}$ cannot continue for arbitrarily large energies because it ultimately violates the Froissart bound, which states that hadronic total cross-sections at asymptotically high energies cannot grow faster than $\ln^2 s$, but this is an upper bound, not a limit. Experimental data for the $pp$ cross-section at LHC and air shower energies show a behaviour that is compatible with a saturated Froissart bound, but whether this is true for _all_ hadronic (and photohadronic) cross-sections or whether it is just a coincidence of the currently available data that they can all be fitted with a $\ln^2 s$ parametrization, is unclear.

To make a long story short, having different high-energy extrapolations available is a good idea for the time being IMO.

After looking in the literature (in particular L. Morejon et al JCAP 11 (2019) 007 [arXiv:1904.07999]), the region of resonances looks different for nuclei than for nucleons; while nucleons have several distinct resonances (apart from the three prominent ones above, SOPHIA contains yet six more), in nuclei the nucleon resonances are smeared out into one broad peak, which is to good accuracy universal for all nuclei. Moreover, this broad resonance appears to be unaffected by shadowing, i.e. $\sigma_{A \gamma} = A \sigma_\text{univ}$ at low energies in contrast to the high energy behaviour which can be written as  $\sigma_{A \gamma} = A^\alpha \sigma_\text{univ}$ with $\alpha < 1$ a function of photon energy.

While it would be interesting to implement this in PROPOSAL, that goes significantly beyond the scope of this PR.

asandrock (approved):
I am unsure about the naming of the parametrizations, but cannot think of something better for now.

Jean1995:
Thanks @asandrock for the investigation!

The only remaining question for me would be the question of the default parametrization of the photoproduction class.
Currently, this is `Rhode`. If I remember currently, the only reason I've chosen this parametrization as a default is because it takes the resonance region into account. 
Should we keep this as a default, or choose another one?

PR #355, opened 2023-02-22
Pin conan to >=1.33, < 2

maxnoe:


Jean1995:
This is the best solution until I've updated the conan package to be compatible with conan 2.0.
But I don't know how long this will take, so this can be merged.

Jean1995 (approved):

PR #365, opened 2023-04-20
Update repository for conan2

Jean1995:
This PR updates the conanfile, the installation instructions, setup.py, and the github actions scripts to be compatible with conan2.

The old installation approach, which relied on the `cmake_paths` generator of conan, has been deprecated.
Instead, I have oriented myself on the procedure presented in the [conan2 tutorial](https://docs.conan.io/2/tutorial/consuming_packages.html), which uses the approach where conan creates a CMake toolchain file which tells CMake where to find the packages installed by conan. This has the advantage that the conan installation process is completely separated from the rest of our CMake structure.

The installation procedure using conan and CMake now has a few more steps, but is more transparent for the user.

`pip install proposal` still works ous of the box. With the update to conan2, a lot of issues that required manual intervention by us (for example the `libcxx` settings) have been resolved, which also simplified the `setup.py`.

The description of the installation process as well as the `setup.py` are _not_ backwards-compatible with conan 1.x.  
Therefore I gave `pip install "conan~=2.0"` as an explicit instruction in the INSTALL.md for now.

### Other changes

- For the python CI, I have added a macOS runner, because in the past, we had problems with the pip installation exclusive to MacOS

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -8,7 +8,7 @@ All different installation approaches are going to be explained in the following
 
 For more detailed information about the specific building tools, see the listed documentations: 
 
-- [conan documentation](https://docs.conan.io/en/latest/)
+- [conan documentation](https://docs.conan.io/2/)
 - [CMake documentation](https://cmake.org/cmake/help/latest/)
 - [pip documentation](https://pip.pypa.io/en/stable/)
 


maxnoe:
Might this be a good point to also try to get rid of the custom cmake code in setup.py and try if we can use `scikit-build` or even `scikit-build-core`?

conanfile.py
```diff
@@ -1,76 +1,128 @@
...
+        self.requires("cubicinterpolation/0.1.5", transitive_headers=True, transitive_libs=True)
+        # spdlog: requires transitive_libs due to direct calls to functionality from headers
+        self.requires("spdlog/1.11.0", transitive_headers=True, transitive_libs=True)
+        # nlohmann_json: public headers include json.hpp and json_fwd.hpp
+        self.requires("nlohmann_json/3.11.2", transitive_headers=True)
         if self.options.with_python:
             self.requires("pybind11/2.10.1")
         if self.options.with_testing:
-            self.requires("boost/1.78.0")
+            self.requires("boost/1.75.0")
```
sudojan, line 75: why the downgrading from 1.78 to 1.75

Jean1995: `cubicinterpolation` itself has a requirement of 1.75.0, and somehow this creates a conflict in conan when proposal defined a different version.
However, this is only relevant if one wants to build the tests. It shouldn't make any practical difference.

Jean1995:
> Might this be a good point to also try to get rid of the custom cmake code in setup.py and try if we can use `scikit-build` or even `scikit-build-core`?

Since we are touching the `setup.py` anyway, now would be the right time if we want to improve it. However, I've never worked with the `scikit-build` generator so far.

INSTALL.md
```diff
@@ -18,45 +18,46 @@ For more detailed information about the specific building tools, see the listed
...
+Now make and install PROPOSAL
+
+```sh             
+$ cmake --build . -j4
+$ cmake --install .
+```
 
+Per default, CMake installs PROPOSAL to `/usr/local`. To change this, specify the install directory when calling CMake, for example `cmake .. -DCMAKE_TOOLCHAIN_FILE=conan_toolchain.cmake -DCMAKE_INSTALL_PREFIX=/path/to/dir`.
 
 *Note:* As an alternative, you may create a local conan package and use it in your project. See the [conan documentation](https://docs.conan.io/en/latest/) for more information.
```
The-Ludwig, line 62:
```suggestion
*Note:* As an alternative, you may create a local conan package and use it in your project. See the [conan documentation](https://docs.conan.io/2/) for more information.
```
Probably applies to more places. `docs.conan.io/en/latest` does not link to the latest conan docs, but to the latest conan 1.x docs.

conanfile.py
```diff
@@ -1,76 +1,128 @@
...
 import os
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.files import copy, get, rmdir
+from conan.tools.microsoft import is_msvc
+from conan.tools.scm import Version
+
+required_conan_version = ">=1.53.0"
```
The-Ludwig, line 10: Just to be sure: this conanfile is still compatible with conan 1?

The-Ludwig:
> > Might this be a good point to also try to get rid of the custom cmake code in setup.py and try if we can use `scikit-build` or even `scikit-build-core`?
> 
> Since we are touching the `setup.py` anyway, now would be the right time if we want to improve it. However, I've never worked with the `scikit-build` generator so far.

I never worked with compiled parts of a python package, but I wonder how `conan` integrates into `scikit-build`? Wouldn't `PROPOSAL` still need custom `setup.py` code for `conan`? Interesting  thread I found on this: https://github.com/orgs/scikit-build/discussions/645

maxnoe:
Maybe also open an issue about a possible conan integration in the new scikit-build-core project? They are rewriting from scratch and handling the C++ dependency is probably important for more people:

The-Ludwig:
> Maybe also open an issue about a possible conan integration in the new scikit-build-core project? They are rewriting from scratch and handling the C++ dependency is probably important for more people: https://github.com/scikit-build/scikit-build-core/issues

Maybe it's better if you open an issue, since I have no idea about scikit-build and C++ distribution through pip and only used conan in one project :D

conanfile.py
Jean1995: Yes. At the moment, the conanfile is written in such a way that it is backwards-compatible with conan 1.x.

The-Ludwig:
> How to we want to proceed with the `scikit-build`(`-core`) suggestion? Open an issue in their repository asking about conan, and (for now) use the custom cmake solution?

I would do so, unless you we want this to be a pilot-project for a custom conan+scikit-build workflow...

Jean1995:
How to we want to go forward with this MR? This is the last issue I would like to solve for the next release.

Does this concept work for you?

sudojan (approved):

PR #369, opened 2023-06-12
Protect Moliere evaluation from negative x

Jean1995:
When `MoliereInterpol` creates the interpolation tables, it evaluates the function at negative x. 
This becomes a problem if floating point exceptions are caught. This simple patch protects the function from evaluation of the square root at negative x.

### DIFF
--- a/src/PROPOSAL/detail/PROPOSAL/scattering/multiple_scattering/Moliere.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/scattering/multiple_scattering/Moliere.cxx
@@ -284,7 +284,7 @@ double Moliere::F1M(double x)
     for (int p = 68; p >= 0; p--)
         sum = sum * x + c1[p] / (2. * p + 1.);
 
-    return sum * std::sqrt(x);
+    return sum * std::sqrt(std::max(x, 0.));
 }
 
 //----------------------------------------------------------------------------//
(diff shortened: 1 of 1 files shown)


asandrock (approved):
In the future, it might perhaps make sense to replace std::max(x, 0.) by something without a branch condition; but as the argument of sqrt or erf this is definitely not decisive from an optimization point of view.

Jean1995:
> > When MoliereInterpol creates the interpolation tables, it evaluates the function at negative x.
> 
> Why? Would that be avoidable?
> 
> Wouldn't a better fix be avoiding calling the function with improper input instead of coercing all negative values to 0? 
> 
> 

I believe this is the interpolation routine evaluating the function around x=0, probably to calculate the differential. But I can investigate this a bit more.

Jean1995:
> > > When MoliereInterpol creates the interpolation tables, it evaluates the function at negative x.
> > 
> > 
> > Why? Would that be avoidable?
> > Wouldn't a better fix be avoiding calling the function with improper input instead of coercing all negative values to 0?
> 
> I believe this is the interpolation routine evaluating the function around x=0, probably to calculate the differential. But I can investigate this a bit more.

I've looked at this, and indeed this comes from the fact that the cubic interpolation calculates the derivative at the edges of the definition regime. I am not sure whether this is intrinsic to the cubic interpolation algorithm. But if it is a bug, this is a bug of the cubic interpolation package.

Issue #382, opened 2023-12-06
Electron-positron pair production: Numerical instable energy asymmetry sampling


Jean1995:
For electron-positron pair production by photons: When sampling the energy asymmetry `x = E_electron / (E_electron + E_positron)`, one would expect a symmetric distribution.

If I sample a distribution for `x`, however, this is not fulfilled. See this example where I sample `x` for a photon energy of 10 MeV:

![image](https://github.com/tudo-astroparticlephysics/PROPOSAL/assets/15159319/7de2ee40-e60a-4193-b937-52fb2727113d)

Same problem for Tsai instead of KochMotz:

![image](https://github.com/tudo-astroparticlephysics/PROPOSAL/assets/15159319/d4edacaf-e94d-44a3-8368-96aae076c5f4)

The issue persists when going to higher energies, here for a photon energy of 1e5 MeV:

![image](https://github.com/tudo-astroparticlephysics/PROPOSAL/assets/15159319/b69e71ce-822c-4c37-9bd1-06b4ebe16343)

![image](https://github.com/tudo-astroparticlephysics/PROPOSAL/assets/15159319/972031f9-d2fd-48ed-b646-d7cd6daffe6d)

In each case, `X Interpol` refers to the secondary calculation using the `CrossSectionInterpol` functions, while `X Integral` refers to calculations based on the `CrossSectionIntegral` functions. The integral calculations seems to be fine, while only the interpol calculations behave incorrectly.

Jean1995:
Related to #349 and #345

Jean1995:
I have openened a PR (#383) where I attempted a fix. This at least fixes the issue for high-energies (where pairproduction is also the dominant process), but it is not able to completely fix the problem at lower energies.

PR #383, opened 2023-12-06
Adapt transformation function for PhotoPairInterpolant

Jean1995:
This is an attempt at solving issue #382 

The first approach that brought better results was to adapt the transformation function in `x` for the interpolation tables.

At the moment, the nodes in `x` in the interpolation tables are distributed linearly in the logarithmic room. However, as x is always between 0 and 1, and symmetric around 0.5, this transformation rule is not really suitable here. 
Therefore, I have changed the notes to be linear in x.

For large energies, this significantly improves the results. For small energies, there are still some problems, although the distribution at least looks more symmetric now.

I have also implemented functions to use the CrossSectionIntegral instead of CrossSectionInterpol functions for the secondary calculators of PhotoPairProduction, see #382, so I am able to make comparisons.

### Improvements

Here are the sampled distributions for the current master and with this PR.
Note that the computation time stays identical.

#### 1e5 MeV Photons:

Old distributions:

image

New distributions:

image

##### 10 MeV photons:

Old distributions:

image

New distributions:

image

##### 2 MeV photons:

Here one can still see a significant deviation between Integral and Interpol. 

Old distributions:

image

New distributions:

image

### DIFF
--- a/src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/PhotoPairProduction.cxx
+++ b/src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/PhotoPairProduction.cxx
@@ -126,7 +126,7 @@ double crosssection::PhotoPairKochMotz::DifferentialCrossSection(
     };
     auto dNdx_nocorrection = i.Integrate(limits.v_min, limits.v_max, integrand, 3);
     auto A = interpolant_->InterpolateArray(comp.GetNucCharge(), energy) / dNdx_nocorrection;
-    return A * DifferentialCrossSectionWithoutA(p, comp, energy, v);
+    return std::max(A * DifferentialCrossSectionWithoutA(p, comp, energy, v), 0.);
 }
 
 double crosssection::PhotoPairKochMotz::DifferentialCrossSectionWithoutA(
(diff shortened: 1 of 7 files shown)


Jean1995:
Attention: For whatever reasons, tests are having problems at the moment...

src/PROPOSAL/detail/PROPOSAL/crosssection/parametrization/PhotoPairProduction.cxx
```diff
@@ -126,7 +126,7 @@ double crosssection::PhotoPairKochMotz::DifferentialCrossSection(
     };
     auto dNdx_nocorrection = i.Integrate(limits.v_min, limits.v_max, integrand, 3);
     auto A = interpolant_->InterpolateArray(comp.GetNucCharge(), energy) / dNdx_nocorrection;
-    return A * DifferentialCrossSectionWithoutA(p, comp, energy, v);
+    return std::max(A * DifferentialCrossSectionWithoutA(p, comp, energy, v), 0.);
```
Jean1995, line 129: This actually requires re-calculation of the A factors! Otherwise, we get a problem...

PR #384, opened 2023-12-06
Fix sources for local conanfile.py to use local files

### LINKED ISSUE #368: `conan export` locally not working
with the new conanfile.py, using conan export locally with conan 1.x doesn't seem to work anymore

Jean1995:
Fixes #368 

Doing something like `conan create . 7.6.3@jean/local01` (with conan 1.x) didn't work because it tried fetching the sources from some external url defined in a conandata.yml. This makes sense for the conanfile.py in the conan-center-index, but not locally in the GitHub repository. Instead, we just want to use the source files within this repository, which is done with `export_sources`.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -35,6 +35,8 @@ class PROPOSALConan(ConanFile):
         "with_documentation": False,
     }
 
+    exports_sources = "*"
+
     @property
     def _min_cppstd(self):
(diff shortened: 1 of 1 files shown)


Jean1995:
this is closed for now since I don't think we should keep supporting conan 1.x

PR #388, opened 2024-02-01
Fix 387

Jean1995:


MaxSac (approved):
Many Thanks for the fast implementation

PR #389, opened 2024-02-09
WIP: use scikit-build-core

maxnoe:
This should get you started.

More metadata can be added and I recommend to also include `setuptools_scm` for automatic version handling, this would require a top-level `proposal/` pure python package though, instead of just shipping the extension library.

This can be made backwards compatible by importing everything from `proposal.....so` in `proposal/__init__.py`.


Jean1995:
If I understand this concept correctly, calling conan and pip install is now separated - which I would prefer to the old solution where conan gets called from within setup.py.

I see from the CI how this can work when locally cloing the repository. But would this also be possible via pypi?
So that one can do something like
```
conan install . -o with_python=True -o with_testing=False --build=missing # optional: fetch dependencies
pip install proposal # install proposal
```
or how would this work?

maxnoe:
> I see from the CI how this can work when locally cloing the repository. But would this also be possible via pypi?

No... because the conanfile is of course only there once you downloaded the package.

This is one of the reasons why this is still "draft". Installing the *source* distribution from pypi would require having all dependencies already on the system.

One solution around this would be to offer pre-built wheels for the most common platforms, e.g. via cibuildwheel.
Then most users will just install the compiled binary, without ever needing conan or even compiling proposal itself.

If we really want to call conan, we could call it from CMakeLists.txt I think, in an `if(SKBUILD AND NOT NO_CONAN)` block or similar

PR #399, opened 2025-05-08
Catch negative variances in ContRandBuilder.h

### LINKED ISSUE #398: Unsafe behavior in continuous randomization possible
I've tried to simulate muons with very low v_cuts (1e-5), and ran into the rather unhelpful error message:

> terminate called after throwing an instance of 'std::logic_error'
  what():  Given rate (nan) for given energy (nan) by drawn random number (0.068190) is larger than the overall crosssection rate (nan).

I looked into what is going on, and realized that there is a problem in the calculation of the continuous randomization:

https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/c3b01b7a7006674c9f83db8ea20f882f7169029b/src/PROPOSAL/PROPOSAL/propagation_utility/ContRandBuilder.h#L29-L39

At an energy (I'll add the entire code for reproducibility at the end) of around `109.11`, the result of `Variance` was negative, which lead to the randomized energy to become `nan`.
 
The underlying reason were problems in the interpolation of the dE2dx function:

Image

**Proposed solution:**

Since this occurs at very low energies close to the particle mass, randomization should not be crucial here.

A simple check such as `auto std = std::sqrt(std::max(Variance(initial_energy, final_energy), 0.0));` should be sufficient. This would set the variance to zero in case of numerical problems, effectively disabling continuous randomization.

Jean1995:
fixes #398 

Simple patch to avoid negative variances, which lead to undefined behavior during propagation.

### DIFF
--- a/src/PROPOSAL/PROPOSAL/propagation_utility/ContRandBuilder.h
+++ b/src/PROPOSAL/PROPOSAL/propagation_utility/ContRandBuilder.h
@@ -29,7 +29,7 @@ public:
     inline double EnergyRandomize(
         double initial_energy, double final_energy, double rnd, double min_energy = 0) final
     {
-        auto std = std::sqrt(Variance(initial_energy, final_energy));
+        auto std = std::sqrt(std::max(Variance(initial_energy, final_energy), 0.0));
         // the parameter 'min' passed to SampleFromGaussian has been
         // min(disp->GetLowerLim(), final_energy) before min_energy was added
         return SampleFromGaussian(final_energy, std, rnd,


Jean1995:
This is the c++ code example I promised

`// Your First C++ Program

#include <iostream>

using namespace PROPOSAL;


int main() {
    InterpolationSettings::TABLES_PATH = "/home/alam_je/proposal_tables";

    auto p_def = MuMinusDef();
    auto medium = Water();
    auto cuts = std::make_shared<EnergyCutSettings>(INF, 1e-4, true);
    auto cross = GetStdCrossSections(p_def, medium, cuts, true);

    auto collection = PropagationUtility::Collection();
    collection.interaction_calc = make_interaction(cross, true);
    collection.displacement_calc = make_displacement(cross, true);
    collection.time_calc = make_time(cross, p_def, true);
    collection.cont_rand = make_contrand(cross, true);
    collection.scattering = make_scattering(MultipleScatteringType::MoliereInterpol, {}, p_def, medium);

    auto prop_utility = PropagationUtility(collection);

    auto density_distr = std::make_shared<Density_homogeneous>(medium);
    auto world = std::make_shared<Sphere>(Cartesian3D(0, 0, 0), 1e20);

    auto sector = std::make_tuple(world, prop_utility, density_distr);
    std::vector<Sector> sec_vec = {sector};

    auto prop = Propagator(p_def, sec_vec);

    auto init_state = ParticleState();
    init_state.energy = 1000.;
    init_state.position = Cartesian3D(0, 0, 0);
    init_state.direction = Cartesian3D(0, 0, -1);

    prop.Propagate(init_state);

sudojan (approved):
passt:)

PR #400, opened 2025-05-08
Fix CI

Jean1995:
There are apparently some issues with CI that needs to be fixed

### DIFF
--- a/.github/workflows/cpp.yml
+++ b/.github/workflows/cpp.yml
@@ -14,47 +14,12 @@ jobs:
             - os : "macos-latest"
               compiler : "clang"
               compilerpp : "clang++"
-            - os : "ubuntu-20.04"
-              compiler : "gcc-7"
-              compilerpp : "g++-7"
-            - os : "ubuntu-20.04"
-              compiler : "gcc-5"
-              compilerpp : "g++-5"
             - os : "windows-latest"
       env:
         CC: ${{ matrix.compiler }}
         CXX: ${{ matrix.compilerpp }}
       steps:
         - uses: actions/checkout@v4
-        - name: Set up gcc5
-          if : ${{ matrix.compiler == 'gcc-5' }}
-          run: |
-            mkdir ~/Downloads
-            cd ~/Downloads
-            wget -c http://archive.ubuntu.com/ubuntu/pool/universe/g/gcc-5/gcc-5-base_5.5.0-12ubuntu1_amd64.deb
-            wget -c http://archive.ubuntu.com/ubuntu/pool/universe/i/isl-0.18/libisl15_0.18-4_amd64.deb
-            wget -c http://archive.ubuntu.com/ubuntu/pool/universe/g/gcc-5/cpp-5_5.5.0-12ubuntu1_amd64.deb
-            wget -c http://archive.ubuntu.com/ubuntu/pool/universe/g/gcc-5/libasan2_5.5.0-12ubuntu1_amd64.deb
-            wget -c http://archive.ubuntu.com/ubuntu/pool/universe/g/gcc-5/libmpx0_5.5.0-12ubuntu1_amd64.deb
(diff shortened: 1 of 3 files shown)


Jean1995:
The ubuntu-20.04 runner of github-actions has been discontinued. We need to either adapt the gcc5/gcc7 tests to run on a different ubuntu version, or just abandon those tests.

maxnoe:
I'd suggest to drop those ancient compilers. They are not used by any currently supported operating system.

Jean1995:
It looks like the integral tests are failing for EPairProduction on MacOS, but only on Linux.
Which is weird, because this means that the calculated results are different for MacOS and Linux (i.e., clang vs. gcc).

The settings for which we have the failures are:
`Epairproduction.Test_of_dEdx`, `EMinus, uranium, 500.0, 1, 1, 1, 10000000000, SandrockSoedingreksoRhode`
and
`Epairproduction.Test_of_dNdx`, `MuMinus, ice, 500.0, 1, 1, 0, 10000000000000, SandrockSoedingreksoRhode`.

Unfortunately I don't have a MacOS machine at hand right now, so I can't debug this issue right now. Didn't we have such problems already in the past?

maxnoe:
If its really clang vs GCC, that's easy to test without macos, you can just install clang on Linux

asandrock:
So the result is that it is not a question of clang vs. g++, but apparently clang on Linux vs. clang on MacOS?

If it is only a problem for the SandrockSoedingreksoRhode parametrization, then the culprit could probably be a function only used there that is not as numerically stable as it ought to be. The only one springing to mind would be the dilogarithm, and that implementation was adapted from the GSL, if I remember correctly.

Jean1995:
> So the result is that it is not a question of clang vs. g++, but apparently clang on Linux vs. clang on MacOS?
> 
> If it is only a problem for the SandrockSoedingreksoRhode parametrization, then the culprit could probably be a function only used there that is not as numerically stable as it ought to be. The only one springing to mind would be the dilogarithm, and that implementation was adapted from the GSL, if I remember correctly.

Assuming I did everything correct with building PROPOSAL with clang on Linux (which I'd strongly believe), this is either Linux vs. MacOS, or maybe clang vs. apple-clang? This makes debugging trickier for me, because it won't be easy to figure out exactly the problem comes from.

maxnoe:
Might also be libc++ vs. libstdc++, as I think that even when building with clang on linux, it defaults to linking against the gcc standard library.

Jean1995:
> Might also be libc++ vs. libstdc++, as I think that even when building with clang on linux, it defaults to linking against the gcc standard library.

This is true. I've now explicitly built PROPOSAL and the dependencies with 'libc++' instead (and `clang`), but the test are still not failing.

sudojan:
@Jean1995 and I found the issue and it is as you @asandrock guessed the dilog; more specific this line
https://github.com/tudo-astroparticlephysics/PROPOSAL/blob/master/src/PROPOSAL/detail/PROPOSAL/math/MathMethods.cxx#L142
```
return -(S * (B0 - _H * B2) + A);
```
as B0 and _H*B2 are of similar size.
The smaller the input values the larger the differences between MacOS (Apple Clang) and Linux (gcc).
Especially for values of 1e-10 the differences are in the order of 1e-2.
We found an alternative implementation of the dilog
https://arxiv.org/pdf/2201.01678
There, the differences between Linux and Apple are less significant and e.g. for the dNdX integration the value stays the same between both systems.
@asandrock can you check, if the implementation is more precise and could be used in the future. However, this would mean all TestFiles needs to be updated.

sudojan:
A comparison of the old implementation in PROPOSAL of the dilogarithm and https://arxiv.org/abs/2201.01678 shows, that 2201.01678 is more robust especially for small values.
dilagarithm_comparison
Therefore, we will switch to the new algorithm.

